PR #1, opened 2016-06-06
Code Review #1

Half-Shot:


mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+            if (error) {
+                console.log("Error", error);
+                return false;
+            } else {
+                try {
+                    var jsonresponse = JSON.parse(body);
+                    if (jsonresponse.token_type == "bearer") {
+                        fs.writeFile("bearer.tok", jsonresponse.access_token, (err) => {
+                            if (err) {
+                                console.console.error("Couldn't write bearer token to file.");
```
kegsay: `console.console.error`? Also, would be good to log the `err`!

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+                } catch (e) {
+                    console.error("Error getting bearer token:", e);
+                    cb(null);
+                }
+
+            }
+        });
+    }
+
+    this.get_bearer_token((bt) => {
```
kegsay: It's generally regarded "surprising" behaviour if constructing an object has large side-effects like making HTTP requests and writing to files. I would let the caller manage this.

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+        if (bt != false) {
```
kegsay: Errr wut? This works, sure, but it's horrible. Your callback on `get_bearer_token` is explicitly returning either a `string` or `null`, yet we do `!= false` checks. Why not `if (bt != null) {`?

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+var mtwitter = function(bridge, config) {
+    this.app_auth = config.app_auth;
+    this.app_twitter = null;
+    this.tuser_cache = {};
+    this.timeline_list = [];
+    this.timeline_queue = [];
+    this.timeline_period = 0;
+    this.timeline_intervalobj = null;
+    this._bridge = bridge;
+    this.get_bearer_token = function(cb) {
```
kegsay: I would prefer it if your object functions were defined on the prototype rather than the particular instantiation. Whilst it matters very little for this use case as you're not doing inheritance, there are a number of benefits regardless:
- One less indent ^___^
- Clearer separation between constructor and instance functions (not sharing the same variable scopes)
- Clarity that the defined function is _supposed_ to be an instance function (due to the presence of `.prototype`), rather than a random function tacked onto an existing object.

In practice, this means changing this:

``` javascript
var mtwitter = function(bridge, config) {
  ...
  this.get_bearer_token = function(cb) {
    ...
  }
}
```

into:

``` javascript
var mtwitter = function(bridge, config) {
  ...
}

mtwitter.prototype.get_bearer_token = function(cb) {
  ...
}
```

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+            for (var name in this.tuser_cache) {
+                var cached = this.tuser_cache[name];
+                if (ts - cached.cache_time > TWITTER_PROFILE_INTERVAL) {
+                    continue;
+                }
+                if (cached.user != null && cached.user.id_str == id) {
+                    resolve(cached.user);
+                }
+            }
+            var twit = this;
```
kegsay: If you're going to alias `this`, always use a consistent, generic term to avoid confusion (typically we use `self`). Though in practice, you're _already using arrow functions_, so I would remove this entirely and arrow function the closures.

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+            this.app_twitter.get('users/show', {
+                user_id: id
+            }, function(error, user, response) {
+                if (error) {
+                    console.error(error);
+                    reject(error);
+                } else {
+                    twit.tuser_cache[user.screen_name] = {
```
kegsay: Change the function definition on `:90` to `(error, user, response) => {` then `s/twit/this/`

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+                }
+            }
+            var twit = this;
+            this.app_twitter.get('users/show', {
+                user_id: id
+            }, function(error, user, response) {
+                if (error) {
+                    console.error(error);
+                    reject(error);
+                } else {
```
kegsay: Prefer the form:

``` javascript
if (condition) {
  // do stuff
  return;
}
// else stuff
```

rather than:

``` javascript
if (condition) {
  // do stuff
} else {
  // else stuff
}
```

Two reasons for this:
- Less indentation on the `else` block ^____^
- Much easier to grok as your mind can completely discard the `if` block after the `return`. If you _didn't_ do this, you would need to check what happens _after the else block_, which may be really really far away! E.g.
  
  ``` javascript
  function foo(callback) {
    if (something) {
      // do some stuff
      callback("something!")
    } else {
      // do else stuff
      // solve world hunger
      // world peace
      // other problems
      // something else?
      // just need to pad this stuff
      // with something
      // anything
      // I'm not picky
    }
    callback("something else!")
  ```
  
  At a glance, you may not notice that `if (true)` would invoke the `callback` _twice_. This is a contrived example, but has caused many real-world problems!

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+                        user: user
+                    };
+                    resolve(user);
+                }
+            });
+        });
+    }
+
+    this.get_user = function(name) {
+        var twit = this;
```
kegsay: Arrow function and remove this.

twitter-as.js
```diff
@@ -0,0 +1,236 @@
+var http = require("http");
+
+var MTwitter = require("./mtwitter.js").MatrixTwitter;
```
kegsay: Generally, the var name should be the same as the class name unless there is a really good reason not to. This is not a good reason.

twitter-as.js
```diff
@@ -0,0 +1,236 @@
...
+    run: function(port, config) {
+        bridge = new Bridge({
+            homeserverUrl: "http://localhost:8008",
+            domain: "localhost",
+            registration: "twitter-registration.yaml",
+            controller: {
+                onUserQuery: function(queriedUser) {
+                    return new Promise(
+                        (resolve, reject) => {
+                            Twitter.get_user_by_id(queriedUser.localpart.substr(8)).then((tuser) => {
```
kegsay: You should promise chain here rather than wrapping the promise returned by `Twitter.get_user_by_id` in another, unchained, `Promise`. If you want me to explain this further, let me know.

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+                        user: user
+                    };
+                    resolve(user);
+                }
+            });
+        });
+
+    }
+
+    this.enqueue_timeline = (userid, localroom, remoteroom) => {
```
kegsay: What does this actually do? What is its purpose?

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+            "created_at": tweet.created_at,
+            "likes": tweet.favorite_count,
+            "reblogs": tweet.retweet_count,
+            "tweet_id": tweet.id_str,
+            "tags": tweet.entities.hashtags,
+            "msgtype": type
+        }
+    }
+
+    this.process_tweet = (bridge, roomid, tweet, treeN) => {
```
kegsay: What does this actually do? What is its purpose?

mtwitter.js
```diff
@@ -0,0 +1,244 @@
...
+                }
+            });
+            setTimeout(() => {
+                intent.sendEvent(roomid, "m.room.message", tweet_content)
+            }, treeN * 250); //Make sure not to send them too quickly.
+        }
+        //TODO: We need to make sure not to spam the HS 
+    }
+
+    this.process_timeline = () => {
```
kegsay: What does this actually do? What is its purpose?

mtwitter.js
Half-Shot: I have to have a list of twitter users, local and remote rooms to iterate over continuously as I poll for new tweets. During startup and whenever a new timeline is requested, it gets added to the queue. I can't think of a better way to handle it.

kegsay: Putting what you've just said as a succinct comment would be lovely :)

mtwitter.js
```diff
@@ -0,0 +1,254 @@
+var Twitter = require('twitter');
+var Request = require('request');
+var fs = require('fs');
+var Buffer = require('buffer').Buffer;
+
+var TWITTER_PROFILE_INTERVAL = 60000;
```
kegsay: Units please!

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+  this.app_auth = config.app_auth;
+  this.app_twitter = null;
+  this.tuser_cache = {};
+  this.timeline_list = [];
+  this.timeline_queue = [];
+  this.timeline_period = 0;
+  this.timeline_intervalobj = null;
+  this._bridge = bridge;
+
+  this.get_bearer_token((bt) => {
```
kegsay: Why is this still defined in the constructor?

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+          this.app_auth.bearer_token = bt;
+          console.log("Twitter Application Auth OK!");
+          this.app_twitter = new Twitter(this.app_auth);
+          return;
+      }
+      console.error("Twitter Application Failed Auth. The bridge will be operating in a limited capacity.");
+  });
+}
+
+mtwitter.prototype.get_bearer_token = function(cb){
```
kegsay: Document what the callback will be called with please. For the record, a callback with just:

``` javascript
function(token) {
  // ...
}
```

isn't great, because this means all you can say on an error is `token === null`. The de-facto standard for Node.js callbacks is to do:

``` javascript
function(err, token) {
  // ...
}
```

so caller's can do whatever they want on an error, and inspect the error to see if they can recover.

Alternatively, return a `Promise` :)

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+  try {
+      fs.accessSync('bearer.tok', fs.R_OK);
+      var tok = fs.readFileSync('bearer.tok', 'utf8');
```
kegsay: This is lazy and unexpected. Callers are giving you a callback to invoke, and yet you're doing I/O synchronously.

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+      cb(tok);
+      return;
+  } catch (e) {
+      console.log("Bearer token either does not exist or cannot be read, getting a new token.")
+  }
+
+  var key = this.app_auth.consumer_key + ":" + this.app_auth.consumer_secret;
+  key = Buffer.from(this.app_auth.consumer_key + ":" + this.app_auth.consumer_secret, 'ascii').toString('base64');
```
kegsay: Any reason why not just:

``` javascript
key = Buffer.from(key, 'ascii').toString('base64');
```

mtwitter.js
kegsay: Also, missing space before `{` - perhaps you should use a style checker?

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+      headers: {
+          'Authorization': "Basic " + key
+      },
+      form: "grant_type=client_credentials",
+      contentType: "application/x-www-form-urlencoded;charset=UTF-8"
+  };
+  Request.post(options, function(error, response, body) {
+      if (error) {
+          console.log("Error", error);
+          return false;
```
kegsay: AFAIK this callback does not expect a boolean return value, so why are you returning `false` rather than just `return;`?

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+      } else {
```
kegsay: If you're returning on `:52`, you don't need this `else` can can outdent.

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+          try {
+              var jsonresponse = JSON.parse(body);
+              if (jsonresponse.token_type == "bearer") {
+                  fs.writeFile("bearer.tok", jsonresponse.access_token, (err) => {
+                      if (err) {
+                          console.error("Couldn't write bearer token to file. Reason:",err);
+                      }
+                  });
+                  cb(jsonresponse.bearer_token);
```
kegsay: Comments please. It's surprising that you're invoking the callback early rather than waiting for `bearer.tok` to be written to disk first. It's okay to do it like this, but you should explain why (e.g. because if you can't write to file then meh, can always get a new one).

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+      }
+      console.error("Twitter Application Failed Auth. The bridge will be operating in a limited capacity.");
+  });
+}
+
+mtwitter.prototype.get_bearer_token = function(cb){
+  try {
+      fs.accessSync('bearer.tok', fs.R_OK);
+      var tok = fs.readFileSync('bearer.tok', 'utf8');
+      cb(tok);
```
kegsay: How do you know this token hasn't been revoked? Might it be worth checking to see if the token is still valid before giving this to the caller, or do you expect the caller to assert validity?

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+              var jsonresponse = JSON.parse(body);
+              if (jsonresponse.token_type == "bearer") {
+                  fs.writeFile("bearer.tok", jsonresponse.access_token, (err) => {
+                      if (err) {
+                          console.error("Couldn't write bearer token to file. Reason:",err);
+                      }
+                  });
+                  cb(jsonresponse.bearer_token);
+              } else {
+                  console.error("Error getting bearer token: Unexpected response");
```
kegsay: No shame in being more precise with your error messages :)

```
POST /oauth2/token did not return a token_type of 'bearer'
```

mtwitter.js
```diff
@@ -0,0 +1,254 @@
+var Twitter = require('twitter');
+var Request = require('request');
+var fs = require('fs');
+var Buffer = require('buffer').Buffer;
+
+var TWITTER_PROFILE_INTERVAL = 60000;
+
+var mtwitter = function(bridge, config) {
```
kegsay: It bugs me that this isn't capitalised. It's a `Class`, but it's not styled as such.

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+            }
+            if (cached.user != null && cached.user.id_str == id) {
+                resolve(cached.user);
+            }
+        }
+        this.app_twitter.get('users/show', {
+            user_id: id
+        }, (error, user, response) => {
+            if (error) {
+                console.error(error);
```
kegsay: No shame in being more precise with your errors. What would you see if this fired? Something like this perhaps?

```
Connection timed out.
```

Boy that's useful and helps you to debug where everything went wrong! Compare it to this:

```
get_user_by_id: GET /users/show returned: Connection timed out.
```

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+                user: user
+            };
+            resolve(user);
+        });
+    });
+
+}
+
+/*
+  Add a user's timeline to the iterator. Tweets will be automatically send to
```
kegsay: > Add a user's timeline to the iterator

Er, what? I see two lists (no iterators) here. The function name is `enqueue` which makes me think that something is going to `dequeue` it, is that correct? If yes, then state what will happen when it is `dequeue`d. If no, then change the name of this function to be something like `add_timeline`.

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+    };
+    this.timeline_list.push(obj);
+    this.timeline_queue.push(obj);
+}
+
+/*
+  This function will fill the content structure for a new matrix message
+  for a given tweet.
+*/
+mtwitter.prototype.construct_message = function(tweet, type) {
```
kegsay: Good comment, poor function name. `message` is far too vague here. `get_matrix_content` perhaps?

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+    return {
+        "body": tweet.text,
+        "created_at": tweet.created_at,
+        "likes": tweet.favorite_count,
+        "reblogs": tweet.retweet_count,
+        "tweet_id": tweet.id_str,
+        "tags": tweet.entities.hashtags,
```
kegsay, line 199: Do all `tweet`s have an `entities` property? If yes, carry on. If no, then sanity check to avoid throwing an error. If not sure, then sanity check.

Typically, it is wise to have a solid contract for all objects you deal with. You _don't_ want to be in a scenario where you _sometimes_ have safe properties to access (always there) and _sometimes_ don't. That's bad because at any given point in your code, you don't know if you've checked for the existence of that property, so you do it again, and again, and again. It's far nicer to say "if I have a `tweet` object then it will _always_ have these properties. This does mean you might need to default-initialise properties if the REST API you're getting it from is omitting keys.

It's good practice to set these things 'at the edges' (just as you get it from disk/the wire/wherever) of your program so the rest of your functions don't need to care.

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+        "msgtype": type
+    }
+}
+
+/*
+  Process a given tweet (including resolving any parent parent tweets), and
```
kegsay: > parent parent tweets

Typo?

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+  submit it to the appropriate room.
```
kegsay: s/appropriate/given/

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+*/
+mtwitter.prototype.process_tweet = function(bridge, roomid, tweet, treeN) {
```
kegsay: Document what `treeN` is please.

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+    if (tweet.in_reply_to_status_id_str != null) {
+        this.app_twitter.get('statuses/show/' + tweet.in_reply_to_status_id_str, {}, (error, newtweet, response) => {
+            if (!error) {
+                this.process_tweet(bridge, roomid, newtweet, treeN);
+                return;
+            }
+            console.error(error);
+        });
+        setTimeout(() => {
+            intent.sendEvent(roomid, "m.room.message", tweet_content)
```
kegsay: I would inline `tweet_content` so I don't need to juggle that in my mind when reading the rest of this function:

``` javascript
intent.sendEvent(roomid, "m.room.message", this.construct_message(tweet, type))
```

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+        }, treeN * 250); //Make sure not to send them too quickly.
+    }
+}
+
+/*
+  Internal function to process the timeline queue.
+*/
+mtwitter.prototype._process_timeline = function(self) {
```
kegsay: Why are you passing `self` in here? This isn't python. You have an implicit `this`.

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+    if (self.timeline_queue.length < 1) {
+        return;
+    }
+    
+    var tline = self.timeline_queue[0];
+    self.timeline_queue = self.timeline_queue.slice(1);
+    var id = tline.remote.getId().substr(9);
```
kegsay: Whenever you manipulate strings in a non obvious way, you should always add a comment explaining what you _intend_ to do e.g:

``` javascript
// strip "twitter-" off the id
var id = tline.remote.getId().substr(9);
```

What are you intending to do here?

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+/*
+  Internal function to process the timeline queue.
+*/
+mtwitter.prototype._process_timeline = function(self) {
+    if (self.timeline_queue.length < 1) {
+        return;
+    }
+    
+    var tline = self.timeline_queue[0];
+    self.timeline_queue = self.timeline_queue.slice(1);
```
kegsay: You can do this better.

``` javascript
var tline = self.timeline_queue.shift()
```

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+        user_id: id,
+        count: 3
+    };
+    var since = tline.remote.get("twitter_since");
+    if (since != undefined) {
+        req.since_id = since;
+    }
+
+    self.app_twitter.get('statuses/user_timeline', req, (error, feed, response) => {
+        if (!error) {
```
kegsay: Invert this to allow you to outdent to get more space.

``` javascript
if (error) {
  console.error(error);
  return;
}
// everything else
```

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+            if (feed.length > 0) {
+                tline.remote.set("twitter_since", feed[0].id_str, 3);
+                feed = feed.reverse();
+                for (var item in feed) {
```
kegsay: Misleading variable name which makes me think you aren't sure what the `in` operator is doing, even though you use it correctly on the next line. There are _very very few_ valid reasons to use any of the `in` operators (including `for ... in`). It is a horrible operator, and I strongly discourage its use as there are clearer alternatives in ES5 and above.

Have some documentation which details why they are horrible operators:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

What do you do instead then? Simple. `forEach()`:

``` javascript
feed.forEach((item) => {
  self.process_tweet(self._bridge, tline.local.roomId, item, 3);
});
```

mtwitter.js
```diff
@@ -0,0 +1,254 @@
...
+            }
+            return;
+        }
+        console.error(error);
+    });
+    self.timeline_queue.push(tline);
+}
+
+mtwitter.prototype.stop_timeline = function() {
+    if (this.timeline_intervalobj) {
```
kegsay: Technically everything is an object, so stating `obj` is redundant. `setInterval()` returns an `intervalID` so perhaps change this to `timeline_intervalID`.

twitter-as.js
```diff
@@ -0,0 +1,232 @@
...
+    run: function(port, config) {
+        bridge = new Bridge({
+            homeserverUrl: "http://localhost:8008",
+            domain: "localhost",
+            registration: "twitter-registration.yaml",
+            controller: {
+                onUserQuery: function(queriedUser) {
+                    var twitter_user;
+                    var image_uri;
+                    return Twitter.get_user_by_id(queriedUser.localpart.substr(8)).then( (tuser) => {
```
kegsay: State what `queriedUser.localpart.substr(8)` is doing:

``` javascript
// strip off "twitter_"
return Twitter.get_user_by_id(queriedUser.localpart.substr(8)).then( (tuser) => {
```

Alternatively, don't use magic numbers and let V8 optimise it away:

``` javascript
return Twitter.get_user_by_id(queriedUser.localpart.substr("twitter_".length)).then( (tuser) => {
```

twitter-as.js
```diff
@@ -0,0 +1,232 @@
...
+                          name: twitter_user.name + " (@" + twitter_user.screen_name + ")",
+                          url: image_uri,
+                          remote: remoteUser
+                      };
+                      console.log(user);
+                      return user;
+                    }).catch((error) => {
+                        console.error("Couldn't find the user.");
+                        console.error("Reason:", error);
+                    });
```
kegsay: This is a gnarly promise chain which isn't really passing anything down the chain and instead setting stuff on the in-scope variables `twitter_user` and `image_uri`. You can do this better:

``` javascript
onUserQuery: function(queriedUser) {
  return Twitter.get_user_by_id(queriedUser.localpart.substr("twitter_".length)).then( (twitter_user) => {
    return uploadImageFromUrl(bridge, twitter_user.profile_image_url_https, queriedUser.getId()).then((uri) => {
      return {
        name: twitter_user.name + " (@" + twitter_user.screen_name + ")",
        url: uri,
        remote: new RemoteUser(twitter_user.id_str)
      };
    );
  }).catch((err) => {
    console.error("Failed to query user: ", err);
  });
}
```

Alternatively, say goodbye to `Promise` and use something like Bluebird's coroutines which make reading this a lot easier - see https://github.com/matrix-org/matrix-appservice-irc/blob/develop/lib/bridge/IrcBridge.js#L269 for an example (look for `yield`).

twitter-as.js
```diff
@@ -0,0 +1,232 @@
...
+                    }
+                });
+            });
+        })
+    }
+}).run();
+
+function roomQuery(alias, aliasLocalpart) {
+    console.log(aliasLocalpart);
+    return Twitter.get_user(aliasLocalpart.substr(9)).then((tuser) => {
```
kegsay: State your intent with `substr(9)` as per other comments.

mtwitter.js
Half-Shot: -_-

mtwitter.js
Half-Shot, line 199: Yep. Tweets should always fill in `entities` and the keys inside. You could argue validating it anyway but then I might as well start validating all the structures I receive, which feels a bit overkill.

kegsay, line 199: > start validating all the structures I receive, which feels a bit overkill.

It's never overkill to sanity check what you're getting from the wire. It's more lines of code, it's more faff to do, it takes fractionally longer to process messages, but you'll then have some guarantees about the data you're manipulating. I appreciate it _is_ annoying to do though, so I don't mind whether you do or do not. However, you should be under no illusion that doing it is in some way pointless/not worth the effort.

What's more, it becomes clearer to outsiders visiting your code as they can see "oh this takes a tweet, which I can see in file X has these properties", which saves them the effort of googling the Twitter API to find the JSON response format to the call you're making. You do run the risk of redocumenting the Twitter API in your code if you take this to the extreme though, so a balance should be struck. In other languages (e.g. Go), you basically decode the JSON into an instance of a class (aka `struct`) which defines exactly what you **think** you should be getting from the wire. Checking those assumptions is never overkill.

If you _do_ decide to go down this route, a fairly noddy `Tweet` class would serve to at least assert the object form:

``` javascript
class Tweet {
  constructor(tweetJson) {
    tweetJson.entities = tweetJson.entities || []; // default `entities` to an empty array if the property is missing
    tweetJson.favorite_count = tweetJson.favorite_count || 0; // default to 0
    // etc..
    this.data = tweetJson;
  }
}

t = new Tweet(someJsonObj);
// e.g. access via t.data.entities
```

Half-Shot, line 199: Understood.I didn't want to get too sidetracked into writing structures for all the little objects I might take from the Twitter API but I understand how useful it is to see the structures in code so you can have a better idea of what you're likely to get. It's something I certainly will get round to doing, but I didn't want to do it while I'm still shifting bits around.

kegsay, line 199: Absolutely, don't feel you need to do it immediately :)

mtwitter.js
```diff
@@ -0,0 +1,291 @@
+var Twitter = require('twitter');
+var Request = require('request');
+var fs = require('fs');
+var Buffer = require('buffer').Buffer;
+
+const TWITTER_PROFILE_INTERVAL = 60000;//ms;
+
+var MTwitter = function (bridge, config) {
+  this.app_auth = config.app_auth;
```
kegsay, line 12: It seems that you only use `config` to pull out `app_auth`. Perhaps you should just pass in an `app_auth`? If you're planning to add more things to the config in the future, then keep it as `config`, but definitely add JSDoc on this constructor to state what properties you expect the `config` to have.

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+                    fs.writeFile("bearer.tok", jsonresponse.access_token, (err) => {
+                        if (err) {
+                            //This error is unfortunate, but not a failure to retrieve a token so the bridge can run fine.
+                            console.error("Couldn't write bearer token to file. Reason:",err);
+                        }
+                    });
+                    //Not waiting for callback since it is trivial to get a new token, and can be done async
+                    resolve(jsonresponse.bearer_token);
+                } else {
+                    reject("Request to oauth2/post did not return the correct token type ('bearer'). This is weeeird.");
```
kegsay, line 68: Add in the `body` to this error message so in the unlikely event this _does_ fire, we can immediately see what Twitter returned instead.

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+        },
+        form: "grant_type=client_credentials",
+        contentType: "application/x-www-form-urlencoded;charset=UTF-8"
+    };
+    Request.post(options, function(error, response, body) {
+        if (error) {
+            reject(error);
+        } else {
+            try {
+                var jsonresponse = JSON.parse(body);
```
kegsay, line 57: I notice we just immediately parse this response if there isn't an `error` set. So what would happen if Twitter returned a 500? You never check the `response.statusCode` so what would happen if you parsed the `body` of the response? If `body` is a string which doesn't parse as JSON it will throw, and then reject with a rather unhelpful `SyntaxError`, obfuscating the underlying cause which was that Twitter returned a body which wasn't JSON because they returned a 500.

This applies to basically every Twitter API call you're making.

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+      //Test the token
+      var auth = {
+        consumer_key: this.app_auth.consumer_key,
+        consumer_secret: this.app_auth.consumer_secret,
+        bearer_token: content
+      };
+      this.app_twitter = new Twitter(auth).get('application/rate_limit_status', {}, (error,status,response) => {        
+        if(error){
+          console.log("Authentication with existing token failed. ");
+          fs.unlink('bearer.tok', (err) => {
```
kegsay, line 96: You're dropping `err` to the floor here. You should at least log it.

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+      }
+      //Test the token
+      var auth = {
+        consumer_key: this.app_auth.consumer_key,
+        consumer_secret: this.app_auth.consumer_secret,
+        bearer_token: content
+      };
+      this.app_twitter = new Twitter(auth).get('application/rate_limit_status', {}, (error,status,response) => {        
+        if(error){
+          console.log("Authentication with existing token failed. ");
```
kegsay, line 95: You're dropping `error` to the floor here. You should at least log it.

kegsay, line 95: Also, is this _really_ the right check to determine token validity? What if it returned a `4xx` response?

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+        }
+      });
+    });
+  });
+}
+
+MTwitter.prototype.get_user_by_id = function(id) {
+    var ts = new Date().getTime();
+    return new Promise((resolve, reject) => {
+        for (var name in this.tuser_cache) {
```
kegsay: I would rather this used `Object.keys` but this works too.

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+        "tags": tweet.entities.hashtags,
+        "msgtype": type
+    }
+}
+
+//Runs every 500ms to help not overflow the room.
+MTwitter.prototype._process_queue = function(){
+  if(this.msg_queue.length > 0){
+    var msg = this.msg_queue[0];
+    this.msg_queue = this.msg_queue.slice(1);
```
kegsay: Why aren't you using `shift()`?

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+    var intent = this._bridge.getIntent(msg.userId);
+    intent.sendEvent(msg.roomId, msg.type, msg.content);
+  }
+}
+
+/*
+  Process a given tweet (including resolving any parent tweets), and
+  submit it to the given room.
+  TreeN - depth of the
```
kegsay: Depth of what?! TELL MEEEEEE!

twitter-as.js
```diff
@@ -0,0 +1,226 @@
...
+function eventQuery(request, context) {
+    var event = request.getData();
+    console.log(event.type);
+    if (event.type == "m.room.member") {
+        if (event.membership == "invite" && event.state_key.startsWith("@twitter_")) { //We should prolly use a regex
+            var intent = bridge.getIntent(event.state_key);
+            intent.join(event.room_id);
+
+            //Set the avatar based on the 'owners' avatar.
+            intent.getClient().getProfileInfo(event.state_key, 'avatar_url').then((url) => {
```
kegsay: Is `url` always going to be set? If they don't have a profile pic, won't this be `null`?

twitter-as.js
```diff
@@ -0,0 +1,226 @@
...
+                console.log("Set Room Avatar:", url);
+                intent.sendStateEvent(event.room_id, "m.room.avatar", "", {
+                    "url": url.avatar_url
```
kegsay: Ohhh, confusing. `url` is the event content? Will `avatar_url` always be set? :angel:

twitter-as.js
```diff
@@ -0,0 +1,226 @@
...
+                Twitter.add_timeline(event.state_key, context.rooms.matrix, context.rooms.remote);
+            } else {
+                console.log("Couldn't find the remote room for this timeline.");
+            }
+        }
+    }
+
+    //console.log("Request:",request);
+    //console.log("Context:",context);
+    return; // we will handle incoming matrix requests later
```
kegsay: This `return`..... does absolutely _nothing_.

twitter-as.js
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/*
+  This will create a stream room for one users timeline.
```
kegsay: `user's`

twitter-as.js
```diff
@@ -0,0 +1,226 @@
...
+  Users who are not authenticated via OAuth will recieve the default power of 0
+  Users who are authenticated via Oauth will recieve a power level of 10
+  The owner of this stream will recieve a 75
```
kegsay: `receive`

twitter-as.js
```diff
@@ -0,0 +1,226 @@
...
+        invite: [roomOwner],
+        initial_state: [
+            powers, {
+                "type": "m.room.join_rules",
+                "content": {
+                    "join_rule": "public"
+                },
+                "state_key": ""
+            }, {
+                "type": "org.twitter.data",
```
kegsay: Are you Twitter? No, you're not. So please don't use `org.twitter`'s namespace. `org.matrix.twitter.data` would be fine, as would `uk.halfshot.twitter.data`.

twitter-as.js
```diff
@@ -0,0 +1,226 @@
...
+        remote: remote
+    };
+}
+
+//WIP
+function uploadImageFromUrl(bridge, url, id = null, name = null) {
+    var contenttype;
+    return new Promise((resolve, reject) => {
+        https.get((url), (res) => {
+            contenttype = res.headers["content-type"];
```
kegsay, line 205: Even though this function is called `uploadImageFromUrl`, it is a lie. There's nothing specific here to images. If you want it to be generic content that's fine, rename the function. If you only want it to be images, you should do at least _some_ due diligence and check the `contenttype` for `image/`. It would also be nice to make it clear what the ultimate return is (a Promise which resolves to a `mxc://` content URI).

mtwitter.js
```diff
@@ -0,0 +1,291 @@
...
+  this.tuser_cache = {};
+  this.timeline_list = [];
+  this.timeline_queue = [];
+  this.timeline_period = 0;
+  this.timeline_intervalID = null;
+  this.msg_queue = [];
+  this._bridge = bridge;
+
+
+  this.get_bearer_token().then((token) => {
```
kegsay: As I said before, it is slightly surprising when constructing a new object results in big side effects like making HTTP hits, writing to disk and starting a long-running poll. You should probably have a `start()` method for this.

mtwitter.js
```diff
@@ -0,0 +1,291 @@
+var Twitter = require('twitter');
+var Request = require('request');
+var fs = require('fs');
+var Buffer = require('buffer').Buffer;
+
+const TWITTER_PROFILE_INTERVAL = 60000;//ms;
```
kegsay: Nit: `TWITTER_PROFILE_INTERVAL_MS` and drop the comment :)

kegsay:
Getting there! Mostly happy with this :D

mtwitter.js
Half-Shot, line 12: > If you're planning to add more things to the config in the future, then keep it as `config`

That was exactly the plan. I'll include my intentions.

I do want to document everything up in a proper format, but again I'd rather do that when things are more finalised. If it's okay, I'll leave comments for now?

mtwitter.js
Half-Shot, line 57: > This applies to basically every Twitter API call you're making.

I disagree with this for other calls. The node-twitter library will flag errors for [anything other than a 200](https://github.com/desmondmorris/node-twitter/blob/master/lib/twitter.js#L154) so I don't see the reason to check the error myself. This also goes for the JSON being valid, which is also checked.

I agree for this call I should be more careful.

twitter-as.js
Half-Shot: According to [this](https://dev.twitter.com/overview/general/user-profile-images-and-banners), a default URL will be given so it should never be null. Again, I think this is a case of validating the API to make sure the correct content is returned. I'll add a comment because I think this could be potentially confusing.

twitter-as.js
Half-Shot: In theory, it should. No 'owner's of timelines should be without their avatar as per the `userQuery` function resolving avatars before the users are even created. I'll add an error in anyway, because I'm sure something will break when it's put into use. Also should rename url to content ;)

mtwitter.js
Half-Shot: Makes sense...quite a lot of sense. Why did I ever decide to do this...

mtwitter.js
```diff
@@ -0,0 +1,307 @@
...
+  this.app_twitter = null;
+  this.tuser_cache = {};
+  this.timeline_list = [];
+  this.timeline_queue = [];
+  this.timeline_period = 0;
+  this.timeline_intervalID = null;
+  this.msg_queue = [];
+  this._bridge = bridge;
+
+
```
kegsay: Needless new lines.

mtwitter.js
```diff
@@ -0,0 +1,307 @@
...
+        },
+        form: "grant_type=client_credentials",
+        contentType: "application/x-www-form-urlencoded;charset=UTF-8"
+    };
+    Request.post(options, function(error, response, body) {
+        if (error) {
+            reject(error);
+        } else if (response.statusCode !== 200){
+            reject("Response to bearer token request returned non OK")
+            console.log("Body of response:",body);
```
kegsay, line 53: You should really log the actual returned status code here (in addition to the body).

mtwitter.js
```diff
@@ -0,0 +1,307 @@
...
+        "likes": tweet.favorite_count,
+        "reblogs": tweet.retweet_count,
+        "tweet_id": tweet.id_str,
+        "tags": tweet.entities.hashtags,
+        "msgtype": type
+    }
+}
+
+//Runs every 500ms to help not overflow the room.
+MTwitter.prototype._process_queue = function(){
```
kegsay: It would be a tad clearer to call this something like `_process_head_of_queue` to make it clear you're not fully emptying the queue (which is what I would expect with a "process queue" function).

Issue #2, opened 2016-06-09
Direct Messaging


Half-Shot:

Half-Shot:
The code is in, awaiting code review

Half-Shot:
This works well enough. Bonus points is if we can initiate a DM which is entirely possible.

PR #4, opened 2016-07-08
Code Review #2

Half-Shot:


README.md
```diff
@@ -1,20 +1,22 @@
...
 
 # Features
 
 - Current (alpha-like quality)
   - Read and follow a users timeline
-- To be complete
+  - Send tweets to yourself and other users
   - Follow hashtags
-  - Reply to users
+- To be complete
```
kegsay: `completed`

README.md
```diff
@@ -59,6 +59,24 @@ The bridge should authenticate and be ready for use.
 
 # Usage
 
+## Linking your account
```
kegsay: Which account? Twitter or Matrix?

package.json
```diff
@@ -17,7 +17,14 @@
   },
   "homepage": "https://github.com/Half-Shot/matrix-appservice-twitter#readme",
   "dependencies": {
-    "twitter": "^1.3.0",
-    "matrix-js-sdk": "git+ssh://git@github.com/matrix-org/matrix-js-sdk.git#develop"
+    "html-entities": "^1.2.0",
+    "js-yaml": "^3.6.1",
+    "matrix-appservice-bridge": "git+ssh://git@github.com/matrix-org/matrix-appservice-bridge.git#develop",
```
kegsay: This is dangerous. If you need to use an unreleased version, pin it to a GIT SHA instead.

package.json
```diff
@@ -17,7 +17,14 @@
...
+    "matrix-js-sdk": "git+ssh://git@github.com/matrix-org/matrix-js-sdk.git#develop",
```
kegsay: This is dangerous. If you need to use an unreleased version, pin it to a GIT SHA instead.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
+var RemoteRoom = require("matrix-appservice-bridge").RemoteRoom;
+var RemoteUser = require("matrix-appservice-bridge").RemoteUser;
+var Twitter    = require('twitter');
+var log        = require('npmlog');
+var OAuth      = require('oauth');
+var TwitterHandler = require('./TwitterHandler.js').TwitterHandler;
```
kegsay: Style: Some people like doing what you've done here (multi-line alignments for prettiness), and boy is it pretty. Don't do it.

The problem with this is that it adds needless maintenance burden when adding/removing variables which are aligned like this (you didn't even bother indenting `:1 ~ :5` to align it with this!). What's even worse is if you do decide to religiously do this, you obfuscate the diff when you add a new `var` (all lines will be modified).

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+
+
+
+/**
+ * Construct a AccountServices.
```
kegsay: `an`

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+ * This class is a handler for conversation between users and the bridge bot to
+ * link accounts together
+ *
+ * @class
+ * @extends {external:TwitterHandler}
+ *
+ * @param  {matrix-appservice-bridge.Bridge}   bridge
+ * @param  app_auth OAuth authentication information
+ * @param  app_auth.consumer_key Twitter consumer key
+ * @param  app_auth.consumer_secret Twitter consumer secret
```
kegsay: This isn't valid JSDoc. Give types!

``` javascript
/**
 * @param {Object} app_auth OAuth authentication information
 * @param {string} app_auth.consumer_secret Twitter consumer secret
 */
```

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+ * @param  {TwitterDB}       storage
+ * @param  {MatrixTwitter}   twitter
+ */
+var AccountServices = function (bridge, app_auth, storage, twitter) {
+  TwitterHandler.call(this,bridge,null,"service");
```
kegsay: Style: spaces after comma. You should really add this to your ESLint.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+ * @extends {external:TwitterHandler}
+ *
+ * @param  {matrix-appservice-bridge.Bridge}   bridge
+ * @param  app_auth OAuth authentication information
+ * @param  app_auth.consumer_key Twitter consumer key
+ * @param  app_auth.consumer_secret Twitter consumer secret
+ * @param  {TwitterDB}       storage
+ * @param  {MatrixTwitter}   twitter
+ */
+var AccountServices = function (bridge, app_auth, storage, twitter) {
```
kegsay: /me refers you to https://github.com/matrix-org/matrix-appservice-gitter/pull/8#discussion_r69425769

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+    'https://api.twitter.com/oauth/access_token',
+    this._app_auth.consumer_key,
+    this._app_auth.consumer_secret,
+    '1.0A',
+    "oob",
+    'HMAC-SHA1'
+  );
+}
+
+AccountServices.prototype.processInvite = function (event, request, context){
```
kegsay: Spacing again.

```
function (event, request, context){
function(event, request, context) {
```

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+  var intent = this._bridge.getIntent();
```
kegsay, line 47: In general, you should [prefer composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). There is rarely a good reason to use inheritance in JS proper, and the lack of static typing makes base functions/variables like this surprising to find. Where was this sucker defined? CTRL+F in this file and seemingly nowhere! YAY! My next resolution step would be to check the constructor, see you're calling `TwitterHandler` then check in that file to see if it is there. It is, so YAY! All good, right? Right? Well yes, sorta, but you're now tethered to the _private_ variables defined in _another file_. That's one of the problems with inheritance: children end up knowing far more about their parents than they really should do!

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+  var event = request.getData();
+  intent.join(event.room_id).then( () => {
+    var rroom = new RemoteRoom("service_"+event.sender);
+    rroom.set("twitter_type", "service");
+    this._bridge.getRoomStore().linkRooms(context.rooms.matrix,rroom);
```
kegsay: Again with the spaces after `,`. Not going to mention this again else the signal to noise ratio is going to be bad.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+    intent.sendMessage(event.room_id,{"body":"This bot can help you link/unlink your twitter account with the Matrix Twitter Bridge. If this was not your intention, please kick the bot.","msgtype":"m.text"});
+    //Add the room to the list.
+  });
+};
+
+AccountServices.prototype.processMessage = function (event, request, context){
+  var intent = this._bridge.getIntent();
+  if(event.content.body == "link account"){
+    log.info("Handler.AccountServices",event.sender + " is requesting a twitter account link.");
```
kegsay: Given we're using ES6 arrow functions, we should make liberal use of [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) as well:

``` javascript
log.info(`Handler.AccountServices ${event.sender} is requesting a twitter account link.`);
```

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+      intent.sendMessage(event.room_id,{"body":"We are unable to process your request at this time.","msgtype":"m.text"});
+    });
+  }
+  else if(event.content.body == "unlink account"){
+    this._storage.remove_client_data(event.sender);
+    this._storage.remove_timeline_room(event.sender);
+    this._twitter.detach_user_stream(event.sender);
+    intent.sendMessage(event.room_id,{"body":"Your account (if it was linked) is now unlinked from Matrix.","msgtype":"m.text"});
+  }
+  else if(isNumber(event.content.body)){
```
kegsay: The first two if statements neatly explain what they're doing based on the command entered. This one doesn't. You should say what this is supposed to be other than a number (e.g. inline comment saying `// auth PIN`)

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+    this._storage.get_client_data(event.sender).then((client_data) => {
+      var pin = event.content.body;
+      log.info("Handler.AccountServices","User sent a pin in to auth with.");
```
kegsay: Not a helpful log message. Which user? Is it safe to log the PIN? If yes, log that too.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+            return;
+          }
+          this._storage.set_client_data(id,profile.id,client_data);
+          resolve(profile);
+        });
+      });
+    }
+    else {
+      reject("User has no associated token request data");
+    }
```
kegsay: Flip the conditional then outdent the if statement e.g:

``` javascript
if (!client_data || client_data.oauth_token == "" || client_data.oauth_secret == "") {
  reject("User has no associated token request data");
  return;
}
// The block from :93 ~ :114
```

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+    });
+  }
+}
+
+AccountServices.prototype._oauth_getAccessToken = function (pin,client_data,id) {
+  return new Promise((resolve,reject) => {
+    if(client_data && client_data.oauth_token != "" && client_data.oauth_secret != ""){
+      this._oauth.getOAuthAccessToken(client_data.oauth_token, client_data.oauth_secret, pin, (error,access_token,access_token_secret) =>{
+        if(error){
+          reject(error.statusCode + ": " + error.data);
```
kegsay, line 178: You almost certainly forgot a `return` here.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+        client_data.access_token_secret = access_token_secret;
+        client = new Twitter({
+          consumer_key: this._app_auth.consumer_key,
+          consumer_secret: this._app_auth.consumer_secret,
+          access_token_key: client_data.access_token,
+          access_token_secret: client_data.access_token_secret
+        });
+        client.get("account/verify_credentials",(error,profile) =>{
+          if(error){
+            log.error("Handler.AccountServices","We couldn't authenticate with the supplied access token for "+ id +". Look into this.");
```
kegsay: Useless log line. The `error` variable is _right there_ buddy! **Log it**.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+    else {
+      reject("User has no associated token request data");
+    }
+  });
+}
+
+AccountServices.prototype._oauth_getUrl = function(event,id){
+  return new Promise((resolve,reject) => {
+    this._oauth.getOAuthRequestToken({"x_auth_access_type":"dm"},(error, oAuthToken, oAuthTokenSecret, results) => {
```
kegsay: I don't know what `{"x_auth_access_type":"dm"}` is specifying - and most likely anyone reading this won't either. A comment to the API reference would make things nicer for readers.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+      if(error){
+        reject(["Couldn't get token for user.\n%s",error]);
```
kegsay: - Missing `return`
- Reject with an error please. Dunno what the heck an array is doing here...

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+        });
+      });
+    }
+    else {
+      reject("User has no associated token request data");
+    }
+  });
+}
+
+AccountServices.prototype._oauth_getUrl = function(event,id){
```
kegsay: Document what your Promise resolves/rejects to please.

src/AccountServices.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+  else if(event.content.body == "unlink account"){
+    this._storage.remove_client_data(event.sender);
+    this._storage.remove_timeline_room(event.sender);
+    this._twitter.detach_user_stream(event.sender);
+    intent.sendMessage(event.room_id,{"body":"Your account (if it was linked) is now unlinked from Matrix.","msgtype":"m.text"});
+  }
+  else if(isNumber(event.content.body)){
+    this._storage.get_client_data(event.sender).then((client_data) => {
+      var pin = event.content.body;
```
kegsay: Your validation checks are somewhat lacking. `event.content.body` has been through the `isNumber()` check which is a good start, but this is Javascript. This passes your `isNumber` check:

```
> s = "    53"
'    53'
> !isNaN(parseFloat(s)) && isFinite(s)
true
```

Further, you keep saying it is a PIN (so I assume an integer), yet there are no assertions that it is in fact an integer. This is complicated because you equally don't want to typecast completely because leading 0s will be stripped then. Given you're doing no arithmetic on the PIN, I would be tempted to just `[0-9]` regex this.

src/DirectMessageHandler.js
```diff
@@ -0,0 +1,33 @@
...
+var DirectMessage = function (bridge, twitter) {
+  TwitterHandler.call(this,bridge);
+  this.twitter = twitter;
+}
+
+DirectMessage.prototype.processMessage = function (event, request, context) {
+    if(event.content.msgtype == "m.text"){
+      this.twitter.send_dm(event.sender,event.room_id,event.content.body);
+    }
+}
```
kegsay, line 25: Mixing 2-space and 4-space indentation? :(

src/DirectMessageHandler.js
```diff
@@ -0,0 +1,33 @@
+var log  = require('npmlog');
+var RemoteRoom = require("matrix-appservice-bridge").RemoteRoom;
+var TwitterHandler = require('./TwitterHandler.js').TwitterHandler;
+
+/**
+ * DirectMessage - Handler for direct messages sent from users to the appservice
```
kegsay: Is this name confusing? How would you model a single `DirectMessage`?

src/DirectMessageHandler.js
```diff
@@ -0,0 +1,33 @@
...
+
+DirectMessage.prototype.processMessage = function (event, request, context) {
+    if(event.content.msgtype == "m.text"){
+      this.twitter.send_dm(event.sender,event.room_id,event.content.body);
+    }
+}
+
+DirectMessage.prototype.processAliasQuery = function(name){
+  //TODO: Support initiating DMs to users.
+  // We might decide to to this via directly PMing a user.
```
kegsay: `to to this`?

src/DirectMessageHandler.js
```diff
@@ -0,0 +1,33 @@
...
+  log.info("Handler.DirectMessage","Got alias request");
```
kegsay: Useless. Log the `name` variable!

src/DirectMessageHandler.js
```diff
@@ -0,0 +1,33 @@
...
+  console.log(name);
```
kegsay: See, you should just do this on `:26`!

src/DirectMessageHandler.js
```diff
@@ -0,0 +1,33 @@
...
+  return null;
+}
+
+module.exports = {
+    DirectMessageHandler: DirectMessage
```
kegsay: Well this isn't confusing.

src/HashtagHandler.js
```diff
@@ -0,0 +1,60 @@
...
+ * @param  {MatrixTwitter}   twitter
+ * @param  {matrix-appservice-bridge.Bridge}   bridge
+ */
+var HashtagHandler = function (bridge, twitter) {
+  TwitterHandler.call(this,bridge,"#","hashtag");
+  this.twitter = twitter;
+}
+
+HashtagHandler.prototype.onRoomCreated = function (alias,entry){
+  console.log(entry);
```
kegsay: A good tip here: If you're having to add debug logging (I assume this is given it is a `console.log`) then most likely it will also be a useful thing to log properly, so just go ahead and do it.

src/HashtagHandler.js
```diff
@@ -0,0 +1,60 @@
...
+  this.twitter.add_hashtag_feed(
+    entry.remote.getId().substr("hashtag_".length),
+    entry
+  );
+}
+
+HashtagHandler.prototype.processMessage = function (event, request, context) {
+    this.twitter.send_matrix_event_as_tweet(event,context.senders.matrix,context.rooms.remote);
```
kegsay: And this is why spaces after commas are so useful. Good luck reading how many arguments this function takes.

src/HashtagHandler.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+HashtagHandler.prototype.processAliasQuery = function(name){
+  log.info("Handler.Hashtag","Got alias request ''%s'",name);
+  var botID = this._bridge.getBot().getUserId();
+
+  var remote = new RemoteRoom("hashtag_" + name);
+  remote.set("twitter_type", "hashtag");
+
+  opts = {
```
kegsay: Any reason to not just in-line this in the `return` statement?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+ * @param  config. Bridge configuration (currently only make use of the auth
+*  information)
+ * @param  config.app_auth.consumer_key Twitter consumer key
+ * @param  config.app_auth.consumer_secret Twitter consumer secret
+ * @param  {TwitterDB}       storage
+ */
+var MatrixTwitter = function (bridge, config, storage) {
+  this.app_auth = config.app_auth;
+  this.app_twitter = null;
+  this.tclients = {};
```
kegsay: Don't worry, [I'm like this with everyone](https://github.com/matrix-org/matrix-appservice-gitter/pull/8#discussion_r69431126).

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  this.timeline_queue = [];
+  this.timeline_period = 3050; //Twitter allows 300 calls per 15 minute (We add 50 milliseconds for a little safety).
+  this.timeline_intervalID = null;
+
+  this.hashtag_period = 2050; //Twitter allows 450 calls per 15 minute (We add 50 milliseconds for a little safety).
+  this.hashtag_intervalID = null;
+  this.hashtag_list = [];
+  this.hashtag_queue = [];
+
+  this.user_streams = {};
```
kegsay: https://github.com/Half-Shot/matrix-appservice-twitter/pull/4/files#r70231338

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+
+  this.processed_tweets = new ProcessedTweetList(256,32);  //This will contain all the tweet IDs of things we don't want to repeat.
+  this.sent_dms = new ProcessedTweetList(1,1);  //This will contain the body of the DM posted to a room to avoid reposting it.
+  this.msg_queue = [];
+  this.msg_queue_intervalID = null;
+  this._bridge = bridge;
+  this.tweet_event_cache = {};
```
kegsay: https://github.com/Half-Shot/matrix-appservice-twitter/pull/4/files#r70231338

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+
+  this.hashtag_period = 2050; //Twitter allows 450 calls per 15 minute (We add 50 milliseconds for a little safety).
+  this.hashtag_intervalID = null;
+  this.hashtag_list = [];
+  this.hashtag_queue = [];
+
+  this.user_streams = {};
+
+  this.processed_tweets = new ProcessedTweetList(256,32);  //This will contain all the tweet IDs of things we don't want to repeat.
+  this.sent_dms = new ProcessedTweetList(1,1);  //This will contain the body of the DM posted to a room to avoid reposting it.
```
kegsay: How does this handle sending the exact same message twice?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+ * authenticates the application with Twitter.
+ * @function
+ * @return {Promise}  A promise that returns when authentication succeeds.
+ */
+MatrixTwitter.prototype.start = function(){
+  return this._get_bearer_token().then((token) => {
+    this.app_auth.bearer_token = token;
+    log.info('Twitter','Retrieved token');
+    this.app_twitter = new Twitter(this.app_auth);
+    this.msg_queue_intervalID = setInterval(() => {this._process_head_of_msg_queue();}, TWITTER_MSG_QUEUE_INTERVAL_MS);
```
kegsay: Typically if you're going to have a public API which starts a polling routine, you should guard said public API to protect against dumb callers who call the same thing twice. Up to you if you want to ignore subsequent `start()` requests, `log.warn` them or `throw`, but I insist you do _something_ because trust me, they are really subtle annoying things to debug should it happen.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+        form: "grant_type=client_credentials",
+        contentType: "application/x-www-form-urlencoded;charset=UTF-8"
+    };
+    Request.post(options, function(error, response, body) {
+        if (error) {
+            reject(error);
+        } else if (response.statusCode !== 200){
+            reject("Response to bearer token request returned non OK")
+            console.log("Body of response:",body);
+            console.log("Statuscode of respnse:",response.statusCode);
```
kegsay: I'm surprised you're not using your nifty `log` library here.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+            log.error("Twitter",error);
+            reject("Unexpected response to application/rate_limit_status during bearer token validation. Bailing.");
+        }
+      });
+    });
+  });
+}
+
+MatrixTwitter.prototype._get_intent = function(id){
+  return this._bridge.getIntent("@twitter_" + id + ":" + this._bridge.opts.domain);
```
kegsay: `Bridge.getIntentFromLocalpart` is A Thing and does something similar FYI.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+
+MatrixTwitter.prototype._update_user_timeline_profile = function(profile){
+  var ts = new Date().getTime();
+  return this.storage.get_profile_by_id(profile.id).then((old)=>{
+    var name = true;
+    var avatar = true;
+    if(old != null){
+      name = (old.profile.name != profile.name) || (old.profile.screen_name != profile.screen_name);
+      avatar = (old.profile.profile_image_url_https != profile.profile_image_url_https);
+
```
kegsay: Variable names are misleading: I am foolishly led to believe that `name` and `avatar` are strings rather than "should I set..." (especially given `||` can be used as default assignments for exactly these kinds of use cases). Rename them to something clearer please. Secondly, is `old.profile` _always_ going to exist? You `null` guard `old` but not `.profile`. Finally, whilst I appreciate you're trying to be succinct with `:170`, you're just being confusing, and may even have your check wrong as a result. Look what you're doing on `:177` - you're setting the display name to a string that uses _both_ the `name` and `screen_name`. What happens if `name` is `undefined`? You're going to set the literal string "undefined" in people's display names :(

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  return this.storage.get_client_data(sender).then((creds) => {
+    return new Promise( (resolve,reject) => {
+      if(creds == null){
+          reject("No twitter account linked.");
+      }
+
+      var ts = new Date().getTime();
+      var client;
+      var id = creds.user_id;
+      if(this.tclients.hasOwnProperty(id)){
```
kegsay: You may at this point just want to use an ES6 `Map`.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  client.last_auth = ts;
+  return client;
+}
+
+MatrixTwitter.prototype._get_twitter_client = function(sender) {
+  //Check if we have the account in the cache
+  return this.storage.get_client_data(sender).then((creds) => {
+    return new Promise( (resolve,reject) => {
+      if(creds == null){
+          reject("No twitter account linked.");
```
kegsay: Missing `return` surely?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+      }
+      else {
+        var client = this._create_twitter_client(creds);
+        this.tclients[id] = client;
+        client.get("account/verify_credentials",(error,profile) => {
+          if(error){
+            delete this.tclients[id];//Invalidate it
+            log.error("Twitter","We couldn't authenticate with the supplied access token for " + id + ". Look into this.");
+            reject("Twitter account could not be authenticated.");
+            //TODO: Possibly find a way to get another key.
```
kegsay, line 290: Missing `return`.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+ * @param  {number} id Twitter Id
+ * @return {Promise<TwitterProfile>} A promise containing the Twitter profile
+ * to be returned. See https://dev.twitter.com/rest/reference/get/users/show
+ */
+MatrixTwitter.prototype.get_profile_by_id = function(id) {
+  log.info("Twitter","Looking up T" + id);
+  var ts = new Date().getTime();
+  return this.storage.get_profile_by_id(id).then((profile)=>{
+    if(profile != null){
+      if(ts - profile.timestamp < TWITTER_PROFILE_INTERVAL_MS){
```
kegsay: Will there _always_ be a `timestamp`?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+    return this._get_user({screen_name:name});
+  });
+}
+
+/**
+ * MatrixTwitter.prototype.tweet_to_matrix_content - This function will fill
+ * the content structure for a new matrix message for a given tweet.
+ *
+ * @param  {type} tweet The tweet object from the Twitter API
+ * @param  {type} type  The 'msgtype' of a message.
```
kegsay, line 367: There's a type called `type`?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+}
+
+
+MatrixTwitter.prototype._push_to_msg_queue = function(muser,roomid,tweet,type) {
+  var time = Date.parse(tweet.created_at);
+  var newmsg = {
+    userId:muser,
+    roomId:roomid,
+    time:time,
+    type:"m.room.message",
```
kegsay: Yet more inconsistent spacing, this time no spaces after `:` :(

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  });
+}
+
+/*
+
+  roomid -
+  depth - The maximum depth of the tweet chain (replies to replies) to be traversed.
+  Set this to how deep you wish to traverse and it will be decreased when the
+  function calls itself.
+*/
```
kegsay: Lone comment?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+
+
+/**
+ * MatrixTwitter.prototype.process_tweet - Process a given tweet (including
+ * resolving any parent tweets), and submit it to the given room. This function
+ * is recursive, limited to the depth set.
+ *
+ * @param  {type} roomid Matrix Room ID of the room that we are processing.
+ * @param  {type} tweet  description
+ * @param  {type} depth  The maximum depth of the tweet chain (replies to
```
kegsay: More of those type `types`!

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+ * replies) to be traversed. Set this to how deep you wish to traverse and it
+ * will be decreased when the function calls itself.
+ * @return {Promise[]]}  A promise that resolves once the tweet has been queued.
+ */
+MatrixTwitter.prototype.process_tweet = function(roomid, tweet, depth) {
+    depth--;
+    var muser = "@twitter_" + tweet.user.id_str + ":" + this._bridge.opts.domain;
+    var intent = this._bridge.getIntent(muser);
```
kegsay: `getIntentByLocalpart` may be useful here.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+                        resolve();
+                  });
+                }
+                else{
+                  resolve();
+                }
+            }
+            else
+            {
+                log.error("process_tweet: GET /statuses/show returned: " + error);
```
kegsay: Ewwww, you're going to return a promise which will **never** be fulfilled should this HTTP hit fail?! Why aren't you `reject()`ing it like you do elsewhere?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+}
+
+/**
+ * MatrixTwitter.prototype.send_matrix_event_as_tweet - Takes a message event
+ * from a room and tries to identify the sender and the correct format before
+ * processing it in {@see send_tweet_to_timeline}.
+ *
+ * @param  {type}               event Matrix event data
+ * @param  {extenal:MatrixUser} user  The user who sent the event.
+ * @param  {extenal:RemoteRoom} room  The remote room that got the message.
```
kegsay: You probably meant `external`.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+      }
+    }
+
+    if(extras !== undefined){
+      if(extras.hasOwnProperty("media")){
+        status.media_ids = extras.media.join(',');
+      }
+    }
+
+    status.status = status.status.substr(0,140);
```
kegsay: This is an interesting one. You're modifying the body sometimes to include a hashtag/mention, which is fine. This may however push the status beyond 140 characters, resulting in truncation where truncation was not expected. This is bad UX. Should you be more strict perhaps and outright reject to send the tweet? Up to you how you want to handle this.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+/**
+ * MatrixTwitter.prototype.remove_timeline - Remove a timeline from being
+ * processed.
+ *
+ * @param  {type} userid The Matrix user_id of the timeline owner. This would
+ *                       be a bridge user.
+ */
+MatrixTwitter.prototype.remove_timeline = function(userid){
+  const tlfind = (tline) => { tline.user_id == userid };
+  this.timeline_list  = this.timeline_list.splice(this.timeline_list.findIndex(tlfind),1);
```
kegsay: What does `this.timeline_list` actually **do**? You add and remove from it, but seemingly never read from it?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+ */
+var MatrixTwitter = function (bridge, config, storage) {
+  this.app_auth = config.app_auth;
+  this.app_twitter = null;
+  this.tclients = {};
+  this.storage = storage;
+
+  this.timeline_list = [];
+  this.timeline_queue = [];
+  this.timeline_period = 3050; //Twitter allows 300 calls per 15 minute (We add 50 milliseconds for a little safety).
```
kegsay: Why isn't this a `const SHOUTY_CAPS_VAR`?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  this.timeline_intervalID = null;
+
+  this.hashtag_period = 2050; //Twitter allows 450 calls per 15 minute (We add 50 milliseconds for a little safety).
```
kegsay: https://github.com/Half-Shot/matrix-appservice-twitter/pull/4/files#r70252196

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+
+
+/**
+ * MatrixTwitter.prototype.remove_hashtag_feed - Remove a hashtag so that it
+ * will no longer be processed.
+ *
+ * @param  {string} hashtag The hasgtag to remove (without the #).
+ */
+MatrixTwitter.prototype.remove_hashtag_feed = function(hashtag){
+  const htfind = (feed) => { feed.hashtag == hashtag };
```
kegsay: You never use this function.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+    this.hashtag_queue.push(obj);
+    log.info('Twitter',"Added Hashtag Feed: %s",hashtag);
+}
+
+
+/**
+ * MatrixTwitter.prototype.remove_hashtag_feed - Remove a hashtag so that it
+ * will no longer be processed.
+ *
+ * @param  {string} hashtag The hasgtag to remove (without the #).
```
kegsay: `hasgtag`

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+        clearInterval(this.hashtag_intervalID);
+        this.hashtag_intervalID = null;
+    }
+}
+
+
+/**
+ * MatrixTwitter.prototype.add_hashtag_feed - Add a hashtag to be processed.
+ *
+ * @param  {string} hashtag The hasgtag to add (without the #).
```
kegsay: `hasgtag`

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+}
+
+MatrixTwitter.prototype._process_hashtag_feed = function(){
+  if (this.hashtag_queue.length < 1) {
+      return;
+  }
+
+  var feed = this.hashtag_queue.shift();
+  var req = {
+    q: "%23"+feed.hashtag,
```
kegsay, line 775: You sure `feed.hashtag` needs no escaping?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  }
+
+  this.app_twitter.get('search/tweets', req, (error, results, response) => {
+      if(error){
+        log.error("Twitter","_process_timeline: GET /statuses/user_timeline returned: %s", error);
+        return;
+      }
+      if(results.statuses.length > 0){
+        feed.entry.remote.set("twitter_since", results.search_metadata.max_id_str);
+        this._bridge.getRoomStore().upsertEntry(feed.entry);
```
kegsay: Dropping errors to the floor.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+/**
+ * MatrixTwitter.prototype.attach_user_stream - Start reading live updates from
+ * a Twitter User Stream.
+ *
+ * @param  {string} user The user's matrix ID.
+ * @return {Promise}   A Promise that will resolve with the operation completes.
+ */
+MatrixTwitter.prototype.attach_user_stream = function(user){
+  if(this.user_streams.hasOwnProperty(user)){
+    log.warn("Twitter","Not attaching stream since we already have one connected!");
```
kegsay: You don't `return` here, this log line suggests you probably want to.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  }
+  return this._get_twitter_client(user).then((c) => {
+    var stream = c.stream('user', {with: "followings"});
+    log.info("Twitter","Attached stream for " + user);
+    stream.on('data',  (data) => {
+      if(data.hasOwnProperty("direct_message")){
+        this._process_incoming_dm(data.direct_message);
+      }
+      else if (data.hasOwnProperty("warning")){
```
kegsay: When you're accessing known properties like this, just do `data.warning` rather than `hasOwnProperty`.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+ * a matrix user's personal timeline, it will be created here.
+ *
+ * @param  {string} user The user's matrix ID.
+ * @param  {object}      The user's Twitter profile.
+ * @return {Promise}     A promise that returns once the operation has completed.
+ */
+MatrixTwitter.prototype.create_user_timeline = function(user,profile){
+  //Check if room exists.
+  return this.storage.get_timeline_room(user).then(value =>{
+    if(value == null){
```
kegsay: Invert and return early then outdent.

``` javascript
if (value != null) {
  return;
}
// .. rest of code
```

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+        this._bridge.getRoomStore().linkRooms(mroom,rroom);
+        this.storage.set_timeline_room(user,room.room_id);
+      });
+    }
+  });
+
+}
+
+MatrixTwitter.prototype._push_to_user_timeline = function(user,msg){
+  this.storage.get_timeline_room(user).then(value =>{
```
kegsay, line 908: `value` is a very unhelpful variable name. `room` perhaps?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+  log.info("Twitter.DM","Creating a new room for DMs from %s(%s) => %s(%s)",msg.sender_id_str,msg.sender_screen_name,msg.recipient_id_str,msg.recipient_screen_name);
+  return Promise.all([
+    this.storage.get_matrixid_from_twitterid(msg.sender_id_str),
+    this.storage.get_matrixid_from_twitterid(msg.recipient_id_str)
+  ]).then(user_ids =>{
+    var invitees = new Set([
+      "@twitter_" + msg.recipient_id_str + ":" + this._bridge.opts.domain
+    ]);
+    for(var user_id of user_ids){
+      if(user_id != null){stop_user_stream
```
kegsay: Uhhh....

src/ProcessedTweetList.js
```diff
@@ -0,0 +1,33 @@
...
+ * per room, not the total size of all the rooms.
+ *
+ * @class
+ * @param  {number} cacheSize How many messages to store before cleanup occurs
+ * @param  {number} sliceSize On cleanup, how many messages will be kept
+ */
+var ProcessedTweetList = function(cacheSize,sliceSize){
+    this._internalArray = {};
+    this._cacheSize = cacheSize || 128;
+    this._sliceSize = sliceSize || 16;
```
kegsay: Variable names aren't the most informative. This isn't a generic class, so don't use generic names. You could transplant this into another class and have it make complete sense. I appreciate you've commented what these things do, which is great, but I think you can go further here:
- `_internalArray` - It's not an array. It's a map of `room_id => Array<TweetID>`. `_roomIdToTweetIds` would be clearer.
- `_cacheSize` - Cache of what? Your comment says "How many messages to store before cleanup occurs" - so why not something like `maxStoredTweets`?
- `_sliceSize` - Slice of what? Your comment says "On cleanup, how many messages will be kept" - so why not something like `minStoredTweets`?

src/ProcessedTweetList.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * ProcessedTweetList - Stores tweets/dms by room_id so we don't accidentally
+ * repeat any messages. Please note that the cache and slice size values ar
```
kegsay: `ar`

src/ProcessedTweetList.js
```diff
@@ -0,0 +1,33 @@
...
+ * per room, not the total size of all the rooms.
+ *
+ * @class
+ * @param  {number} cacheSize How many messages to store before cleanup occurs
+ * @param  {number} sliceSize On cleanup, how many messages will be kept
```
kegsay: Would be good to document the default values too.

src/ProcessedTweetList.js
```diff
@@ -0,0 +1,33 @@
...
+}
+
+ProcessedTweetList.prototype.contains = function(room_id,tweet_id){
+  if(!this._internalArray.hasOwnProperty(room_id)){
+    return false;
+  }
+  return (this._internalArray[room_id].indexOf(tweet_id) != -1);
+}
+
+module.exports = ProcessedTweetList;
```
kegsay, line 33: This entire class feels wrong. It isn't really **doing** much. What it _is_ doing is completely generic (except for maybe the room ID lookup). Either make this actually generic, or delegate more responsibilities to this class to make it worthy of being called a `ProcessedTweetList`.

src/TimelineHandler.js
```diff
@@ -0,0 +1,139 @@
...
+}
+
+
+TimelineHandler.prototype.onRoomCreated = function(alias,entry){
+    var owner = entry.remote.data.twitter_user;
+    var intent = this._bridge.getIntent(owner);
+    intent.getClient().getProfileInfo(owner, 'avatar_url').then((content) =>
+    {
+        if (typeof content.avatar_url != "string")
+        {
```
kegsay: What's with the completely different `if`and `function` bracket style?

src/TimelineHandler.js
```diff
@@ -0,0 +1,139 @@
...
+            log.error("Handler.Timeline","User", owner, "does not have an avatar set. This is unexpected.");
```
kegsay: Is it really unexpected?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,987 @@
...
+      }
+    });
+  }).catch(reason =>{
+    log.error("Twitter.DM","Failed to send DM: %s", reason);
+  });
+}
+
+module.exports = {
+    MatrixTwitter: MatrixTwitter
+}
```
kegsay, line 1055: When a file starts approaching 1000 lines.... you know you should probably look at it and see if it can be broken down into more manageable chunks. It's been horrible reviewing this file because there's so many functions that are flying around the place which can be called. You have a nice split between "hashtag" and "timeline" processing: that might be a good partition point.

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+
+/**
+ * TwitterDB - Stores data for specific users and data not specific to rooms.
+ * @class
+ * @param  {string} filepath Location of the SQLite database file.
+ */
+var TwitterDB = function(filepath){
+  this.db = new SQLite3.Database(filepath,(err) => {
+    if(err){
+      log.error("TwitDB","Error opening database, %s");
```
kegsay: This is one of those rare cases where you probably _do_ want to crash the process if this fails.

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+ * @function
+ * @param  {type} id          Twitter ID of the profile.
+ * @param  {string} name      screenname of the profile.
+ * @param  {object} data      The profile data.
+ * @param  {number} timestamp The time when this data was *fetched*.
+ */
+TwitterDB.prototype.set_twitter_profile = function(id,name,data,timestamp){
+  this.db.run(
+    `
+    REPLACE INTO user_cache VALUES ($id,$name,$data,$timestamp);
```
kegsay: So a function called `set_twitter_profile` modifies a table called `user_cache`? Confusing. Also, cache is pretty redundant - we know you're always going to have a stale version from Twitter, so calling it "cache" just makes me think that you're expiring it somehow (which AFAIK you aren't).

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+  ,{
+    $id: id,
+    $name: name,
+    $data: JSON.stringify(data),
+    $timestamp: timestamp
+  },
+  function (err) {
+    if(err){
+      log.error("TwitDB","Error storing profile: %s",err);
+      return;
```
kegsay, line 154: So no one calling `set_twitter_profile` needs to care if it was actually written to the database?

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+    }
+  });
+}
+
+TwitterDB.prototype.get_client_data = function(user_id){
+  return new Promise((resolve,reject) =>{
+    this.db.get(
+      `
+      SELECT *
+      FROM twitter_account
```
kegsay, line 164: Terminology. A function called `get_client_data` I would expect to pull stuff from a `client_data` or maybe `clients` table, but no, it's pulling from `twitter_account`, _obviously_.

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+      }
+    });
+  });
+}
+
+TwitterDB.prototype.set_client_data = function(user_id,twitter_id,data){
+  this.db.run(
+    `
+    REPLACE INTO twitter_account VALUES (
+      $user_id,
```
kegsay: Be **very very careful** relying on column ordering... this is not a great idea if you ever want to modify the table... being explicit is always better.

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+    $oauth_secret: data.oauth_secret,
+    $access_token: data.access_token,
+    $access_token_secret: data.access_token_secret
+  },
+  function (err) {
+    if(err){
+      log.error("TwitDB","Error storing client data: %s",err);
+      return;
+    }
+    log.info("TwitDB","Stored client data for %s",user_id);
```
kegsay: So no caller ever needs to care if `set_client_data` even worked?

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+}
+
+
+TwitterDB.prototype.close = function() {
+  this.db.close();
+}
+
+module.exports = {
+  TwitterDB: TwitterDB
+}
```
kegsay, line 443: Your setters in general are missing return Promises - Some parts of the bridge _do care_ if the data they tried to persist actually worked or not.

src/TwitterDB.js
```diff
@@ -0,0 +1,462 @@
...
+    if(err){
+      log.error("TwitDB","Error storing dm room: %s",err);
+      return;
+    }
+    log.info("TwitDB","Stored dm room %s",room_id);
+  });
+}
+
+//Caches every user profile we grab from Twitter so as to not go over our limits.
+TwitterDB.prototype._create_profile_cache = function(){
```
kegsay: All of these `_create_xxx` functions would benefit from being a single `_create` function which accept an array of strings which contain the `CREATE TABLE` statements. This means that:
- All your CREATE TABLE stuff is in one place rather than spread over many functions.
- You don't repeat yourself when handling errors.

src/TwitterHandler.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param  {external:RoomBridgeStore.Entry} entry description
+ */
+TwitterHandler.prototype.onRoomCreated = function(alias,entry){
+  log.warn("Handler","STUB onRoomCreated");
+  return;
+}
+
+module.exports = {
+    TwitterHandler: TwitterHandler
+}
```
kegsay: Whilst I appreciate wanting to have a common interface here, I'm not sure how much good it is doing. Reading the derived classes, it's not clear which functions were specified in this base class, and all the documentation now resides here as well. You're also assigning properties to each derived handler which then get used thoroughly in each derived class, and it's confusing to try to identify where it was set.

Interfaces are great when you don't know at call-time which actual implementation you want to use, but in your case you really do know at call-time which implementation you want: you call them directly in `TwitterRoomHandler`. This removes any real value in sharing the API between the handlers, so I don't really know what your aim is here.

src/TwitterRoomHandler.js
```diff
@@ -0,0 +1,109 @@
...
+var TwitterRoomHandler = function (bridge, handlers) {
+  this._bridge = bridge;
+  this.handlers = handlers; // 'service' handler
+}
+
+TwitterRoomHandler.prototype.processInvite = function (event,request, context){
+  var remote = context.rooms.remote;
+  if(remote == null
+     && event.sender != "@twitbot:localhost"
+     && event.state_key == "@twitbot:localhost")
```
kegsay: `localhost` references should be nuked. We want people to run this on actual HSes remember?

src/TwitterRoomHandler.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+TwitterRoomHandler.prototype.onRoomCreated = function(alias,roomId){
+  var roomstore = this._bridge.getRoomStore();
+  roomstore.getEntriesByMatrixId(roomId).then(entries =>{
+    console.log(entries);
+    if(entries.length == 0){
+      log.error("RoomHandler","Got a onRoomCreated, but no remote is associated.");
+    }
```
kegsay, line 115: Missing `return`.

src/util.js
```diff
@@ -0,0 +1,124 @@
...
+    if(url.startsWith("https")){
+      var ht = https;
+    }
+    else {
+      var ht = http;
+    }
+
+    ht.get((url), (res) => {
+        var size = parseInt(res.headers["content-length"]);
+        var buffer = Buffer.alloc(size);
```
kegsay: > Added in: Node v5.10.0

You _really_ sure you want to set your min version to 5+?

kegsay: Also, what happens if your response is `Content-Length: foo`?

kegsay:
I'm concerned about your error handling. There were >3 cases where you error check and then fail to return after executing error handling code. You may be at a point where some unit tests would be most useful.

src/AccountServices.js
Half-Shot: I'm probably going to split these three into separate functions with a nice jsdoc description for each, as I think that would look cleaner.

src/AccountServices.js
Half-Shot: I've been spoiled by other languages... I never thought of this.

src/MatrixTwitter.js
Half-Shot: This is a problem where Twitter will often send me tweets through the _very fast_ user streams before response comes back from sending tweets. Therefore, I can't use the ID property to identify them. 

This usually isn't a problem because 
1. Twitter disallows sending the exact same message twice in tweets.
2. As far as direct messages are concerned, I'm only interested in not echoing back what the user just said.

However, I believe there is an unspecified amount of time before a duplicate tweet can be sent which means this could indeed break. The only alternative I can think of is a timestamp+body to identify messages.

src/MatrixTwitter.js
Half-Shot: Yes..but I guess it can't hurt to validate.

src/MatrixTwitter.js
Half-Shot: Hmm fair point. My main concern is if I reject it, how do I inform the user that it could not be sent. In my mind having a truncated message is better than no message if you can't explain to the user why their message failed. 

I _am_ relying on the user knowing the Twitter 140 char limit however. The alternative is sending a public message into the room from the bot saying "oy, keep your messages inside the limit" which isn't ideal.

src/MatrixTwitter.js
Half-Shot: ...Seems I may have used it before but it no longer serves a purpose.

src/MatrixTwitter.js
Half-Shot, line 775: It should be just alphanumeric, and I _should_ be validating that before we get this far.

src/MatrixTwitter.js
Half-Shot: Hmm? Which bit do you mean?

src/MatrixTwitter.js
Half-Shot: _glares at the middle click pasting 'feature'_

src/TwitterDB.js
Half-Shot, line 154: Not really. The system is only used to cache profile information so we don't over use Twitters API and get locked out. If it doesn't get written for any reason, then it just means that one profile won't be cached which isn't a huge deal unless it starts to happen frequently.

src/TwitterDB.js
Half-Shot: The profiles _are_ automatically expired if they become too old, though granted MatrixTwitter.js isn't the place to do that so I would say it is certainly a cache. `set_twitter_profile` makes it sound permanent however, so that needs correcting.

src/TwitterRoomHandler.js
Half-Shot: Not sure how that got in there...

src/util.js
Half-Shot: > You _really_ sure you want to set your min version to 5+?

``` js
if(Buffer.alloc) {//Since 5.10
   buffer = Buffer.alloc(size);
}
else {//Deprecated
   buffer = new Buffer(size);
}
```

How about that?

kegsay: Something like that sounds good to me.

src/MatrixTwitter.js
kegsay: `upsertEntry` returns a Promise, which you aren't doing anything with if it rejects.

src/MatrixTwitter.js
kegsay: You need _some_ form of feedback mechanism - what if the OAuth token has expired and you need the user to do something? A 1:1 room with the bot may suffice.

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * @class
+ * @extends {external:TwitterHandler}
+ * @param  {object} opts
+ * @param  {matrix-appservice-bridge.Bridge}   opts.bridge
+ * @param  {TwitterDB}       opts.storage
+ * @param  {MatrixTwitter}   opts.twitter
+ * @param  {object} opts.app_auth OAuth authentication information
+ * @param  {string} app_auth.consumer_key Twitter consumer key
+ * @param  {string} app_auth.consumer_secret Twitter consumer secret
```
kegsay: This needs to be `opts.app_auth.xxxxx`

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * processInvite - Handler for invites from a matrix user to a (presumably)
+ * empty room. This will join the room and send some help text.
+ * @param  {MatrixEvent} event   The event data of the request.
+ * @param  {Request} request The request itself.
+ * @param  {Context} context Context given by the appservice.
+ */
+AccountServices.prototype.processInvite = function (event, request, context) {
+  log.info("Handler.AccountServices", "Got invite");
```
kegsay: Log the room ID and user who invited the bot.

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+      }
+    );
+    //Add the room to the list.
+  }).catch(err => {
+    log.error("Handler.AccountServices", "Couldn't join service room. %s", err)
+  });
+};
+
+AccountServices.prototype.processLeave = function (event, request, context) {
+  log.info("Handler.AccountServices", "User %s left room. Leaving", event.sender);
```
kegsay: Log the room ID too.

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+};
+
+
+/**
+ * _beginLinkAccount - Processes a users request to link their twitter account
+ * with the bridge. This should return a authorisation link.
+ *
+ * @param  {object} event The Matrix Event from the requesting user.
+ */
+AccountServices.prototype._beginLinkAccount = function (event) {
```
kegsay: You can tidy up the function signature here. You only ever use `event.sender` and `event.room_id`, so why not just accept those as parameters? This would make it clearer exactly what is needed, rather than accepting any arbitrary event.

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+  });
+}
+
+/**
+ * _unlinkAccount - Processes a users request to unlink their twitter account
+ * from the bridge.
+ *
+ * @param  {object} event The Matrix Event from the requesting user.
+ */
+AccountServices.prototype._unlinkAccount = function (event) {
```
kegsay: https://github.com/Half-Shot/matrix-appservice-twitter/pull/4/files#r71678699

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+AccountServices.prototype._oauth_getAccessToken = function (pin, client_data, id) {
+  return new Promise((resolve, reject) => {
+    if (!client_data || client_data.oauth_token == "" || client_data.oauth_secret == "") {
+      reject("User has no associated token request data");
+      return;
+    }
+    this._oauth.getOAuthAccessToken(
+      client_data.oauth_token,
+      client_data.oauth_secret,
+      pin,
```
kegsay: Tempted to put this on the same line as `:174` and then just put the function on a new line, looots of vertical space otherwise.

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+          })
+        });
+      });
+  });
+};
+
+/**
+ * _oauth_getUrl - Start the process of connecting an account to Twitter.
+ *
+ * @param  {type} id    The matrix user id wishing to authenticate.
```
kegsay: Presumably `{string}`

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+ * @return {Promise<string>}  A promise that will return an auth url or reject with nothing.
```
kegsay: rejects with an `Error`.

src/AccountServices.js
```diff
@@ -0,0 +1,257 @@
...
+        if(error) {
+          reject(error);
+        }
+        var data = {
+          oauth_token: oAuthToken,
+          oauth_secret: oAuthTokenSecret,
+          access_token: "",
+          access_token_secret: ""
+        };
+        this._storage.set_twitter_account(id, "", data).then(()=>{
```
kegsay, line 232: Why are we calling this with an empty string for the twitter ID? Do we not know it at the moment?

src/HashtagHandler.js
```diff
@@ -0,0 +1,87 @@
...
+    );
+}
+
+
+/**
+ * processAliasQuery - A request to this handler to provision a room for the
+ * given name *after* the global alias prefix.
+ *
+ * @param  {type} name The requested name *after* '#twitter_'
+ * @return {ProvisionedRoom | Promise<ProvisionedRoom,Error>, null}
```
kegsay: You don't ever return a Promise, so change this to:

```
@return {?ProvisionedRoom}
```

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+/**
+ * MatrixTwitter.prototype.start - Starts the timers for polling the API and
+ * authenticates the application with Twitter.
+ * @function
+ * @return {Promise}  A promise that returns when authentication succeeds.
+ */
+MatrixTwitter.prototype.start = function () {
+  if(this.app_twitter != null) {
+    log.warn("Twitter",  "Attempted to call start() while having been started previously.");
+    return
```
kegsay: You can race here if you call `start()` twice in a row because you don't immediately non-nullify `this.app_twitter` on this turn of the event loop. If you guard using the promise itself, you'll actually protect yourself from calling `start()` twice.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+    this.msg_queue_intervalID = setInterval(() => {
+      this._process_head_of_msg_queue();
+    }, TWITTER_MSG_QUEUE_INTERVAL_MS);
+
+    this.start_timeline();
+    this.start_hashtag();
+
+  }).catch((error) => {
+    log.error('Twitter', 'Error trying to retrieve bearer token:', error);
+    throw "Couldn't get a bearer token for Twitter AS.";
```
kegsay: Throw errors, not strings.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+      } else if (response.statusCode !== 200) {
+        reject("Response to bearer token request returned non OK")
+        log.error("Twitter",
+              "Body of response:%s\nStatuscode of respnse:%s",
+              body,
+              response.statusCode
+            );
+      } else {
+        try {
+          var jsonresponse = JSON.parse(body);
```
kegsay, line 111: The try/catch only needs to be around this line, so you can outdent the rest of this else block and save some space.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+              if (err) {
+                            //This error is unfortunate, but not a failure to retrieve a token so the bridge can run fine.
+                log.error("Twitter", "Couldn't write bearer token to file. Reason:", err);
+              }
+            });
+                    //Not waiting for callback since it is trivial to get a new token, and can be done async
+            resolve(jsonresponse.bearer_token);
+          } else {
+            reject("Request to oauth2/post did not return the correct" +
+                    "token type ('bearer'). This is weeeird.");
```
kegsay: Reject with errrors, not strings.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+  });
+}
+
+MatrixTwitter.prototype._get_bearer_token = function () {
+  return new Promise((resolve) => {
+    fs.readFile('bearer.tok', {encoding: 'utf-8'}, (err, content) => {
+      if(err) {
+        log.warn('Twitter', "Token file not found or unreadable. Requesting new token.");
+        log.error("Twitter", err);
+        resolve(Promise.resolve(this._get_bearer_http()));
```
kegsay: You don't need to wrap `this._get_bearer_http()` in `Promise.resolve` as it already returns a promise.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+
+MatrixTwitter.prototype._get_intent = function (id) {
+  return this._bridge.getIntentFromLocalpart("twitter_" + id);
+}
+
+MatrixTwitter.prototype._update_user_timeline_profile = function (profile) {
+  var ts = new Date().getTime();
+  if(profile == null) {
+    log.warn("Twitter", "Tried to preform a profile update with a null profile.");
+    return;
```
kegsay, line 187: The function signature to `_update_user_timeline_profile` is horrible. You conditionally return a `Promise` :( either return a Promise always or never. Looking at your usages of this function, I would just not return a `Promise` (you never seem to use it).

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+
+MatrixTwitter.prototype._create_twitter_client = function (creds) {
+  var ts = new Date().getTime();
+  var client = new Twitter({
+    consumer_key: this.app_auth.consumer_key,
+    consumer_secret: this.app_auth.consumer_secret,
+    access_token_key: creds.access_token,
+    access_token_secret: creds.access_token_secret
+  });
+  client.last_auth = ts;
```
kegsay, line 239: Adding random properties onto objects you don't control is risky, you don't know if you're going to clobber something that was there. In this case it's probably fine, but you should **definitely** comment why you're doing this.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+        }
+
+        log.info("Twitter", "Credentials for %s need to be reevaluated.", sender);
+        client.get("account/verify_credentials", (error, profile) => {
+          if(error) {
+            log.info("Twitter", "Credentials for " + id + " are no longer valid.");
+            log.error("Twitter", error);
+            delete this.tclients[id];//Invalidate it
+            resolve(this._get_twitter_client(sender));
+          }
```
kegsay, line 270: You are missing a `return`

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+        this.tclients[id] = client;
+        client.get("account/verify_credentials", (error, profile) => {
+          if(error) {
+            delete this.tclients[id];//Invalidate it
+            log.error(
+              "Twitter",
+              "We couldn't authenticate with the supplied access token for %s. Look into this.",
+              id
+            );
+            reject("Twitter account could not be authenticated.");
```
kegsay: You should really be logging the `error` _somewhere_ so you can at least debug stuff!

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+      }
+    });
+  });
+}
+
+MatrixTwitter.prototype._get_user = function (data) {
+  return new Promise((resolve, reject) => {
+    this.app_twitter.get('users/show', data, (error, user) => {
+      if (error) {
+        error = error[0];
```
kegsay: From IRL: The `twitter` library sucks and can sometimes return an array of errors rather than be consistent. We should probably check it is an array to be sure.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+    }
+    return this._get_user({screen_name: name});
+  });
+}
+
+/**
+ * MatrixTwitter.prototype.tweet_to_matrix_content - This function will fill
+ * the content structure for a new matrix message for a given tweet.
+ *
+ * @param  {TwitterTweet} tweet The tweet object from the Twitter API See {@link}
```
kegsay, line 366: See what?!

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+}
+
+//Runs every 500ms to help not overflow the room.
+MatrixTwitter.prototype._process_head_of_msg_queue = function () {
+  if(this.msg_queue.length > 0) {
+    var msgs = this.msg_queue.pop();
+    //log.info("Twitter","Pulling off queue:",msg.content.body);
+    for(var msg of msgs) {
+      var intent = this._bridge.getIntent(msg.userId);
+      intent.sendEvent(msg.roomId, msg.type, msg.content).catch(reason =>{
```
kegsay: This races: there is no guarantee the first `msg` will be sent first (before the second). You should probably promise chain this to make sure you do it in a consistent order.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+  }
+    //log.info("Twitter","Processing tweet:",tweet.text);
+
+  return new Promise( (resolve, reject) => {
+    if (tweet.in_reply_to_status_id_str != null && depth > 0) {
+      this.app_twitter.get(
+            'statuses/show/' + tweet.in_reply_to_status_id_str, {}, (error, newtweet) => {
+              if (!error) {
+                var promise = this.process_tweet(roomid, newtweet, depth)
+                if(promise != null) {
```
kegsay: `process_tweet` will _always_ return a Promise so no need to null guard.

src/AccountServices.js
Half-Shot, line 232: Yes, it's sort of a midway storage to let the bridge know that we have begun the linking process and are waiting for the pin to be returned.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+    log.error("Twitter", "Failed to send tweet. %s", err);
+  });
+}
+
+/*
+  Timeline functions
+*/
+
+MatrixTwitter.prototype._process_timeline = function () {
+  if (this.timeline_queue.length < 1) {
```
kegsay: Also known as:

``` javascript
this.timeline_queue.length === 0
```

Which is clearer.

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+    return;
+  }
+
+
+  var tline = this.timeline_queue.shift();
+  var id = tline.entry.remote.getId().substr("@twitter_".length);
+  var req = {
+    user_id: id,
+    count: 3
```
kegsay: Magic number?

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+  if (since != undefined) {
+    req.since_id = since;
+  }
+
+  this.app_twitter.get('statuses/user_timeline', req, (error, feed) => {
+    if(error) {
+      log.error("Twitter", "_process_timeline: GET /statuses/user_timeline returned: %s", error);
+      return;
+    }
+    if (feed.length > 0) {
```
kegsay: Invert then outdent:

``` javascript
if (feed.length === 0) {
  return;
}
// ... rest of code
```

src/MatrixTwitter.js
```diff
@@ -0,0 +1,1040 @@
...
+
+/**
+ * MatrixTwitter.prototype.remove_hashtag_feed - Remove a hashtag so that it
+ * will no longer be processed.
+ *
+ * @param  {string} hashtag The hashtag to remove (without the #).
+ */
+MatrixTwitter.prototype.remove_hashtag_feed = function (hashtag) {
+  const htfind = (feed) => { feed.hashtag == hashtag };
+  this.hashtag_queue = this.hashtag_queue.splice(this.hashtag_queue.findIndex(htfind), 1);
```
kegsay: Lovely subtle bug here. What if the hashtag is not found? `htfind` returns `false` for everything, making `findIndex` return `-1` by default, and `Array.splice(-1, 1)` pops the last element of the array.

kegsay:
LGTM aside from comments.

src/MatrixTwitter.js
Half-Shot, line 366: The @\see {@\link ...} below?

PR #19, opened 2016-11-14
Implement Provisioning API

Half-Shot:
See doc for information on how it's set up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,10 @@ node_js:
  - "6.9.0"
  - "6.0.0"
 script: "npm test && npm run-script lint"
+notifications:
+    webhooks:
+        urls:
+            - "https://scalar.vector.im/api/neb/services/hooks/dHJhdmlzLWNpLyU0MEhhbGYtU2hvdCUzQWhhbGYtc2hvdC51ay8lMjFOYlRzc3hSVVdSWXRwQk1XZUYlM0FoYWxmLXNob3QudWs"
+        on_success: change  # always|never|change
+        on_failure: always
+        on_start: never
--- a/Provisioning-API.md
+++ b/Provisioning-API.md
@@ -32,6 +32,7 @@ Only used for PUT
 | :------------- | :------------- |
 | 200 | the bridge has updated or deleted the link |
 | 201 | the bridge has created a new link |
+| 400 | One or more parameters were malformed or invalid |
 | 401 | the user is unauthorized to create links |
 | 403 | the bridge could not join the room itself |
 | 404 | the room or the profile wasn't found |
@@ -67,8 +68,8 @@ GET Retrieve links for the given room.
 
 | Return Code    | Reason     |
 | :------------- | :------------- |
-| 200 | the bridge OKd it and 0 or more results exist |
-| 404 | the room  wasn't found |
+| 200 | A set of results exist (which may be empty) |
+| 400 | One or more parameters were malformed or invalid |
 
 
 ```
@@ -88,7 +89,7 @@ GET Retrieve links for the given room.
 ## /_matrix/provision/show/timeline/%screenName%
 ### Methods
 
-GET Retrive some profile information about a timeline.
+GET Retrieve some profile information about a timeline.
 
 ### Parameters
 
@@ -101,6 +102,7 @@ GET Retrive some profile information about a timeline.
 | Return Code    | Reason     |
 | :------------- | :------------- |
 | 200 | The Twitter user was found. |
+| 400 | One or more parameters were malformed or invalid |
 | 404 | The Twitter user wasn't found |
 
 
--- a/Provisioning-API.md
+++ b/Provisioning-API.md
@@ -1,7 +1,7 @@
 All of these API requests can return a 500 if something fails that wasn't expected.
 
-## /_matrix/provision/%roomId%/link/timeline/%screenName%?=%userId
-## /_matrix/provision/%roomId%/link/hashtag/%hashtag%?=%userId
+## /_matrix/provision/%roomId%/link/timeline/%screenName%?userId=%userId
+## /_matrix/provision/%roomId%/link/hashtag/%hashtag%?userId=%userId
 
 PUT - Create a new link
 
--- a/Provisioning-API.md
+++ b/Provisioning-API.md
@@ -2,21 +2,35 @@ All of these API requests can return a 500 if something fails that wasn't expect
 
 ## /_matrix/provision/%roomId%/link/timeline/%screenName%?userId=%userId
 ## /_matrix/provision/%roomId%/link/hashtag/%hashtag%?userId=%userId
+### Methods
 
-PUT - Create a new link
+PUT Create a new link. Can also be used to reconfigure an existing room.
 
-DELETE - Remove an existing link
+DELETE Remove an existing link
+
+### Parameters
 
 | Param          | Description     |
-| :------------- | :------------- |
+| :------------- | :-------------  |
 | roomId | Matrix Room ID |
 | screenName | The screenname of the timeline to add (without @) |
 | hashtag | The hashtag to add (without #) |
 | userId | The user that is requesting the bridge |
 
-| Return Code    | Reason     |
+## Body
+The body is used to pass additional options to the room.
+
+Only used for PUT
+
+| Option          | Description    |
+| :-------------  | :------------- |
+| exclude_replies | Don't show replies to mentions by the user. |
+
+### Return Codes
+
+| Return Code    | Reason         |
 | :------------- | :------------- |
-| 200 | the bridge OKed it or the link already exists |
+| 200 | the bridge OKed it |
 | 401 | the user is unauthorized to create links |
 | 403 | the bridge could not join the room itself |
 | 404 | the room or the profile wasn't found |
@@ -27,11 +41,14 @@ DELETE - Remove an existing link
 }
 ```
 
-## GET /_matrix/provision/%roomId%/links
+## /_matrix/provision/%roomId%/links
+### Methods
 
-Retrieve links for the given room.
+GET Retrieve links for the given room.
 
-| Param          | Description     |
+### Parameters
+
+| Param          | Description    |
 | :------------- | :------------- |
 | roomId | Matrix Room ID |
 
@@ -39,6 +56,8 @@ Retrieve links for the given room.
 | :------------- | :------------- |
 | 200 | the bridge OKd it and 0 or more results exist |
 | 404 | the room  wasn't found |
+
+### Return Codes
 
 ```
 {
@@ -54,14 +73,19 @@ Retrieve links for the given room.
 }
 ```
 
-## GET /_matrix/provision/show/timeline/%screenName%
+## /_matrix/provision/show/timeline/%screenName%
+### Methods
 
-Retrive some profile information about a timeline.
+GET Retrive some profile information about a timeline.
+
+### Parameters
 
 | Return Code    | Reason     |
 | :------------- | :------------- |
 | 200 | the bridge OKd it and 0 or more results exist |
 | 404 | the room  wasn't found |
+
+### Return Codes
 
 | Param          | Description     |
 | :------------- | :------------- |
--- a/Provisioning-API.md
+++ b/Provisioning-API.md
@@ -100,8 +100,8 @@ GET Retrive some profile information about a timeline.
 
 | Return Code    | Reason     |
 | :------------- | :------------- |
-| 200 | the bridge OKd it and 0 or more results exist |
-| 404 | the room  wasn't found |
+| 200 | The Twitter user was found. |
+| 404 | The Twitter user wasn't found |
 
 
 ```
--- a/Provisioning-API.md
+++ b/Provisioning-API.md
@@ -1,0 +1,78 @@
+All of these API requests can return a 500 if something fails that wasn't expected.
+
+## /_matrix/provision/%roomId%/link/timeline/%screenName%?=%userId
+## /_matrix/provision/%roomId%/link/hashtag/%hashtag%?=%userId
+
+PUT - Create a new link
+
+DELETE - Remove an existing link
+
+| Param          | Description     |
+| :------------- | :------------- |
+| roomId | Matrix Room ID |
+| screenName | The screenname of the timeline to add (without @) |
+| hashtag | The hashtag to add (without #) |
+| userId | The user that is requesting the bridge |
+
+| Return Code    | Reason     |
+| :------------- | :------------- |
+| 200 | the bridge OKed it or the link already exists |
+| 401 | the user is unauthorized to create links |
+| 403 | the bridge could not join the room itself |
+| 404 | the room or the profile wasn't found |
+
+```
+{
+  "message": Error message if did not complete successfully.
+}
+```
+
+## GET /_matrix/provision/%roomId%/links
+
+Retrieve links for the given room.
+
+| Param          | Description     |
+| :------------- | :------------- |
+| roomId | Matrix Room ID |
+
+| Return Code    | Reason     |
+| :------------- | :------------- |
+| 200 | the bridge OKd it and 0 or more results exist |
+| 404 | the room  wasn't found |
+
+```
+{
+  "timelines":[
+    {
+      //Profile information (same as /show/timeline)
+    }
+  ]
+  "hasttags":[
+    //List of strings without #
+    "foo", "bar"
+  ]
+}
+```
+
+## GET /_matrix/provision/show/timeline/%screenName%
+
+Retrive some profile information about a timeline.
+
+| Return Code    | Reason     |
+| :------------- | :------------- |
+| 200 | the bridge OKd it and 0 or more results exist |
+| 404 | the room  wasn't found |
+
+| Param          | Description     |
+| :------------- | :------------- |
+| screenName | The screenname of the timeline to add (without @) |
+
+```
+{
+  twitterId:
+  name:
+  screenName:
+  avatarUrl:
+  description:
+}
+```
--- a/Provisioning-API.md
+++ b/Provisioning-API.md
@@ -1,5 +1,3 @@
-All of these API requests can return a 500 if something fails that wasn't expected.
-
 ## /_matrix/provision/%roomId%/link/timeline/%screenName%?userId=%userId
 ## /_matrix/provision/%roomId%/link/hashtag/%hashtag%?userId=%userId
 ### Methods
@@ -19,6 +17,8 @@ DELETE Remove an existing link
 
 ## Body
 The body is used to pass additional options to the room.
+These options can also be used to update the existing room, though currently
+only the timeline can be modified.
 
 Only used for PUT
 
@@ -30,14 +30,25 @@ Only used for PUT
 
 | Return Code    | Reason         |
 | :------------- | :------------- |
-| 200 | the bridge OKed it |
+| 200 | the bridge has updated or deleted the link |
+| 201 | the bridge has created a new link |
 | 401 | the user is unauthorized to create links |
 | 403 | the bridge could not join the room itself |
 | 404 | the room or the profile wasn't found |
 
+#### 200
+
 ```
 {
-  "message": Error message if did not complete successfully.
+  "message":"Link updated" || "Hashtag already bridged!"
+}
+```
+
+#### 201
+
+```
+{
+  "message":"Linked successfully"
 }
 ```
 
@@ -52,12 +63,13 @@ GET Retrieve links for the given room.
 | :------------- | :------------- |
 | roomId | Matrix Room ID |
 
+### Return Codes
+
 | Return Code    | Reason     |
 | :------------- | :------------- |
 | 200 | the bridge OKd it and 0 or more results exist |
 | 404 | the room  wasn't found |
 
-### Return Codes
 
 ```
 {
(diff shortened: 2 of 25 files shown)


config/config.sample.yaml
```diff
@@ -28,3 +28,6 @@ logging:
   count: 3
   compress: false
   level: verbose #One of error,warn,info,verbose,silly,silent
+
+provisioning:
+  enable: true
```
lukebarnard1: Maybe summarise briefly what this does?

Also, there's no mention of `required_power_level` and it's default.

package.json
```diff
@@ -25,7 +25,7 @@
     "html-entities": "^1.2.0",
     "js-yaml": "^3.6.1",
     "logrotate-stream": "^0.2.5",
-    "matrix-appservice-bridge": "^1.3.0",
+    "matrix-appservice-bridge": "matrix-org/matrix-appservice-bridge#develop",
```
lukebarnard1, line 28: Ah yes... We should do a release for mx-as-bridge soon

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+      return has_power;// Detailed error message;
+    }
+
+    // PUT
+    if(createLink) {
+      if (type == "timeline") {
+        return self._linkTimeline(room_id, name);
+      }
+      else if(type == "hashtag") {
+        return self._linkHashtag(room_id, name);
```
lukebarnard1: I'm not convinced this `else if` is necessary, if the coroutine rejects it's promise, will that not just throw? And that surely means that the only possible return values are true/false?

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+    const rooms = yield Promise.filter(roomstore.getEntriesByMatrixId(room_id), item =>{
+      if(item.remote) {
+        if(type == "timeline" && item.remote.data.twitter_type == "timeline") {
+          return self._twitter.get_profile_by_screenname(item.remote.data.twitter_user).then(profile =>{
+            if(!profile) {
+              return false;
+            }
+            return profile.screen_name == name;
+          });
+        }
```
lukebarnard1: I'd be tempted to break DELETE into a separate function, just to make `manageLink` a bit smaller.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+      res.status(500).json({error: "An internal error occured."});
+      log.error("Provisioner", "Error occured: %s", err);
+    }
+  }
+
+  * _manageLink (self, req) {
+    const user_id = req.query.userId;
+    const room_id = req.params.roomId;
+    const type = req.params.type;
+    const name = req.params.name;
```
lukebarnard1: I would be good to validate `type` and `name`, and provide errors if they're malformed, especially as `name` could get put in the DB.

lukebarnard1: Also, some comments to explain what each of these is would be nice.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+        }
+      }
+    });
+
+
+    if(rooms.length == 0) {
+      return {err: 404, body: "Link not found."};
+    }
+
+
```
lukebarnard1: I think it would be clearer to explicitly return false instead of not returning.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+    yield self._bridge.getRoomStore().getEntriesByMatrixId(roomId).then(rooms => {
+      return rooms;
+    }).each(room =>{
+      if(room.remote.data.twitter_type == "timeline") {
+        return self._twitter.get_profile_by_id(room.remote.data.twitter_user).then(profile =>{
+          body.timelines.push({
+            twitterId: profile.id_str,
+            avatarUrl: profile.profile_image_url_https,
+            name: profile.name,
+            screenName: profile.screen_name,
```
lukebarnard1, line 173: Perhaps 

```
let rooms = yield ...
rooms.each( ...
```

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+  // or empty object for hashtags
+  * _queryProfile (self, req) {
+    const profile = yield self._twitter.get_profile_by_screenname(req.params.screenName);
+    if (!profile) {
+      return {err: 404, body: "User not found."}
+    }
+    else{
+      return {
+        twitterId: profile.id_str,
+        avatarUrl: profile.profile_image_url_https,
```
lukebarnard1: Where does the empty object come from?

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+    return this._twitter.get_profile_by_screenname(screenname).then(p =>{
+      if (!p) {
+        return {err: 404, body: "Twitter profile not found!"};
+      }
+      profile = p;
+      return roomstore.getEntriesByRemoteRoomData({
+        twitter_type: "timeline",
+        twitter_user: profile.id_str
+      })
+    }).then(rooms => {
```
lukebarnard1: I'd prefer coroutines over `then` chaining 0:)

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+      this._twitter.timeline.add_timeline(profile.id_str, room_id, true);
+      return {};
+    });
+  }
+
+  _linkHashtag (room_id, hashtag) {
+    const roomstore = this._bridge.getRoomStore();
+    hashtag = hashtag.replace("#", "");
+    return roomstore.getEntriesByRemoteId("hashtag_"+hashtag).then(rooms => {
+      const isLinked = rooms.length > 0;
```
lukebarnard1: `is_new` maybe?

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+      return {};
+    })
+
+
+  }
+
+  isProvisionRequest (req) {
+    return req.url.match(/^\/_matrix\/provision\/(\S+)\/(link|links|timeline)/)
+      || req.url.startsWith("/_matrix/provision/show");
+  }
```
lukebarnard1, line 312: You should validate `hashtag` to make sure it looks like a hastag and then just chop off the first character.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+
+  _updateBridgingState (roomId, userId, status, skey) {
+    const intent = this._bridge.getIntent();
+    try {
+      intent.client.sendStateEvent(roomId, 'm.room.bridging', {
```
lukebarnard1, line 317: I would be good to send back an error code with this, maybe one that isn't 4xx so that a client could show it as informational, rather than a full-blown "An error has occured: ...".

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+    }
+    var powerState;
+    try{
+      powerState = yield matrixClient.getStateEvent(roomId, 'm.room.power_levels');
+    }
+    catch(err) {
+      log.error("Provisioning", `Error retrieving power levels (${err.data.error})`);
+    }
+
+    if (!powerState) {
```
lukebarnard1, line 349: I suggest including `user_id` and `status` in this message somehow.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+      return Promise.reject({err: 403, body: "Couldn't join room. Perhaps room permissions are not set to public?"});
+    }
+    var powerState;
+    try{
+      powerState = yield matrixClient.getStateEvent(roomId, 'm.room.power_levels');
+    }
+    catch(err) {
+      log.error("Provisioning", `Error retrieving power levels (${err.data.error})`);
+    }
+
```
lukebarnard1, line 348: This error is dropped to the floor. It should at least be logged.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+    //Leave if not setup within 10 minutes.
+    setTimeout(() => {
+      self._leaveIfUnprovisioned(roomId);
+    }, LEAVE_UNPROVIS_AFTER_MS);
+
+    let actualPower = 0;
+    if (powerState.users[userId] !== undefined) {
+      actualPower = powerState.users[userId];
+    }
```
lukebarnard1, line 361: Error dropped, please log.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+    else if (powerState.users_default !== undefined) {
+      actualPower = powerState.users_default;
+    }
+
+    let requiredPower = self._config.required_power_level;
+    if (powerState.events["m.room.power_levels"] !== undefined) {
+      requiredPower = powerState.events["m.room.power_levels"]
+    }
```
lukebarnard1, line 373: It's possible (granted, only if there's a bug in mx-js-sdk) that `err` is in fact a type error or something else. I dislike that mx-js-sdk throws anything that isn't an `Error`, because it means you have to check for both. I'd suggest at least null checking `err.data` so you don't get any `Cannot get 'error' of undefined` errors being thrown.

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+      if(result !== undefined) {
+        if(result.err) {
+          res = res.status(result.err);
+        }
+        res.json(result);
+      }
+    }
+    catch (err) {
+      res.status(500).json({error: "An internal error occured."});
+      log.error("Provisioner", "Error occured: %s", err);
```
lukebarnard1: It seems `npmlog` doesn't quite handle this as you'd expect, printing the error out twice. I think the solution would be to remove the '%s', but I'm not sure about the double printing thing.

``` JS
> let log = require('npmlog');
> let err = new Error('I am an error');
> log.error('Some context', "This error has just happened: %s", err);
ERR! Some context Error: I am an error
ERR! Some context     at repl:1:11
ERR! Some context     at sigintHandlersWrap (vm.js:22:35)
ERR! Some context     at sigintHandlersWrap (vm.js:96:12)
ERR! Some context     at ContextifyScript.Script.runInThisContext (vm.js:21:12)
ERR! Some context     at REPLServer.defaultEval (repl.js:313:29)
ERR! Some context     at bound (domain.js:280:14)
ERR! Some context     at REPLServer.runBound [as eval] (domain.js:293:12)
ERR! Some context     at REPLServer.<anonymous> (repl.js:513:10)
ERR! Some context     at emitOne (events.js:101:20)
ERR! Some context     at REPLServer.emit (events.js:188:7)
ERR! Some context  This error has just happened: %s Error: I am an error
ERR! Some context     at repl:1:11
ERR! Some context     at sigintHandlersWrap (vm.js:22:35)
ERR! Some context     at sigintHandlersWrap (vm.js:96:12)
ERR! Some context     at ContextifyScript.Script.runInThisContext (vm.js:21:12)
ERR! Some context     at REPLServer.defaultEval (repl.js:313:29)
ERR! Some context     at bound (domain.js:280:14)
ERR! Some context     at REPLServer.runBound [as eval] (domain.js:293:12)
ERR! Some context     at REPLServer.<anonymous> (repl.js:513:10)
ERR! Some context     at emitOne (events.js:101:20)
ERR! Some context     at REPLServer.emit (events.js:188:7)
```

src/Provisioner.js
```diff
@@ -0,0 +1,341 @@
...
+    if (powerState.events["m.room.power_levels"] !== undefined) {
+      requiredPower = powerState.events["m.room.power_levels"]
+    }
+    else if (powerState.state_default !== undefined) {
+      requiredPower = powerState.state_default;
+    }
+
+    return Promise.resolve(actualPower >= requiredPower);
+  }
+
```
lukebarnard1, line 380: '10 minutes' might change, I suggest naming the delay constant used instead.

src/twitter/Timeline.js
```diff
@@ -149,24 +149,34 @@ class Timeline {
...
+      }
+
+      if(isTimeline) {
+        this._timelines = queue;
       }
+      else{
+        this._hashtags = queue;
+      }
+
     }
```
lukebarnard1, line 219: I think a negation is needed here

src/twitter/Timeline.js
```diff
@@ -149,24 +149,34 @@ class Timeline {
...
           log.warn("Timeline", "Tried to remove %s for %s but it didn't exist", room_id, id);
+          return;
         }
       }
       else {
         queue[i].room = []
       }
+
       if(queue[i].room.length == 0) {
-        queue = this._timelines.splice(i, 1);
```
lukebarnard1, line 195: Negation needed

twitter-as.js
```diff
@@ -1,3 +1,5 @@
+global.Promise = require('bluebird');
```
lukebarnard1: Given than `Promise` is already built-in to the node runtime, it would be clearer to `require` it in each file it is used. The distinction is quite important and it would be useful to know during reviews like this, without going to look for `global.Promise`.

lukebarnard1 (changes requested):
I couldn't leave a comment, but on line 182 of Twitter.js, there's an if statement that uses `url`. There's a potential race here because prior to the if, there's a `then` chain that includes the setting of `url`. This could occur after the if statement.

Also, in general, `===` is best unless you absolutely need the whole convoluted coercion process that occurs with `==`.

Half-Shot:
Seems like git doesn't want to play ball at all. I'll work through this review and then clean it up.

src/twitter/Timeline.js
Half-Shot: It is? I don't see it.

lukebarnard1: Yeah, _not_ sure what I was talking about :sweat_smile:

Half-Shot:
> I couldn't leave a comment, but on line 182 of Twitter.js, there's an if statement that uses url. There's a potential race here because prior to the if, there's a then chain that includes the setting of url. This could occur after the if statement.
> 
> Also, in general, === is best unless you absolutely need the whole convoluted coercion process that occurs with ==.

Good shout. I added it as a rule and there are 85 violations of it soooo that's gonna be fun :laughing:

src/Provisioner.js
Half-Shot: Not sure I follow, I think I return in every possible case here.

.travis.yml
```diff
@@ -3,3 +3,10 @@ node_js:
  - "6.9.0"
  - "6.0.0"
 script: "npm test && npm run-script lint"
+notifications:
```
lukebarnard1: I'm not entirely sure why this has been committed here.

lukebarnard1, line 6: I'm not entirely sure why this has been committed here.

config/config.sample.yaml
```diff
@@ -8,26 +8,26 @@ app_auth: #Get these from twitter by creating an application through the API sit
   consumer_secret: BLAHBLAHBLAH
 
 hashtags:
-  enable: true
-  admin_only: false
+  enable: true # Enable processing hashtag rooms and creating new ones
+  admin_only: false # NOT USED. Only an admin listed can create these rooms.
```
lukebarnard1, line 12: Does `NOT USED` mean deprecated?

config/config.sample.yaml
```diff
@@ -8,26 +8,26 @@ app_auth: #Get these from twitter by creating an application through the API sit
...
+  enable_profile_images: true # Enable download of profile images
 
 logging:
-  file: logs/twitter-bridge.log
-  size: 50m
-  count: 3
-  compress: false
+  file: logs/twitter-bridge.log #Filepath of the current log file. (old ones are rotated.)
+  size: 50m #Maximum size (m,k,b) of a log file before it is rotated.
+  count: 3 # Number of files to keep before rotation.
```
lukebarnard1: Are the files not kept after rotation?

src/twitter/Timeline.js
```diff
@@ -113,38 +121,42 @@ class Timeline {
...
    * @param  {object} opts Options
-   * @param  {boolean} opts.isnew Is the room 'new', and we shouldn't do a full poll.
+   * @param  {boolean} opts.is_new Is the room 'new', and we shouldn't do a full poll.
    * @param  {boolean} opts.exclude_replies Should we not fetch replies.
    */
   add_timeline (twitter_id, room_id, opts) {
     var tline = this._find_timeline(twitter_id);
     var obj;
 
+    if (this.config.timelines.enable) {
```
lukebarnard1: I think this might have been what I was talking about in https://github.com/Half-Shot/matrix-appservice-twitter/pull/19/files/20c6ee2f76bd02bf79f984ce6b183168b579bdfe..6a699dfcdbdd67571603ed8811c0f234544b3bc6#r87995839

Surely if timelines are enabled, you don't want to return.

twitter-as.js
```diff
@@ -130,7 +128,7 @@ var cli = new AppService.Cli({
 
           //Fix rooms that are alias rooms
           // Criteria: canonical_alias is #_twitter_@*+:domain
-          if (type == "timeline" && entry.matrix.get("twitter_user") == undefined) {
+          if (type === "timeline" && entry.matrix.get("twitter_user") == null) {
```
lukebarnard1: Changing `undefined` to `null` doesn't have any effect here, according to the [ES 6 spec:](http://www.ecma-international.org/ecma-262/6.0/#sec-abstract-equality-comparison)
### 7.2.12 Abstract Equality Comparison
```js
3. If Type(x) is the same as Type(y), then
    Return the result of performing Strict Equality Comparison x === y.
```
Which would handle either `undefined == undefined` or `null == null` (`undefined` has type `undefined` and `null` obviously has type `object`, at least with node.js). And in other cases:

```js
...
4. If x is null and y is undefined, return true.
5. If x is undefined and y is null, return true.
```

By changing `undefined` here to `null`, you're swapping from using `4.` in the true case to using `5.`. And, in the false case, if the value is a non-null object it gets compared to `null` because they're both objects, which returns false because of `===`.

Also, by using `==`, you treat `undefined` in the same way as you treat `null`. So long as that is OK in this context, `==` is fine.

lukebarnard1 (changes requested):
A few questions and very few minor changes, but apart from those, LGTM!

src/twitter/Twitter.js
```diff
@@ -29,7 +30,7 @@ class Twitter {
     this._storage = storage;
 
     this._dm = new DirectMessage(this);
-    this._timeline = new Timeline(this);
+    this._timeline = new Timeline(this, this.config.timelines, this.config.hashtags);
```
lukebarnard1: `this.config` -> `this._config`

Provisioning-API.md
```diff
@@ -0,0 +1,102 @@
...
+GET Retrive some profile information about a timeline.
+
+### Parameters
+
+| Return Code    | Reason     |
+| :------------- | :------------- |
+| 200 | the bridge OKd it and 0 or more results exist |
+| 404 | the room  wasn't found |
+
+### Return Codes
```
lukebarnard1, line 100: "Return Codes" and "Parameters" should be swapped

Provisioning-API.md
```diff
@@ -0,0 +1,102 @@
...
+| :------------- | :------------- |
+| screenName | The screenname of the timeline to add (without @) |
+
+```
+{
+  twitterId:
+  name:
+  screenName:
+  avatarUrl:
+  description:
```
lukebarnard1: Some example values would be nice here

Provisioning-API.md
```diff
@@ -0,0 +1,102 @@
...
+## /_matrix/provision/show/timeline/%screenName%
+### Methods
+
+GET Retrive some profile information about a timeline.
+
+### Parameters
+
+| Return Code    | Reason     |
+| :------------- | :------------- |
+| 200 | the bridge OKd it and 0 or more results exist |
```
lukebarnard1: "0 results found" sounds like 404 to me

Provisioning-API.md
```diff
@@ -0,0 +1,102 @@
...
+| Param          | Description    |
+| :------------- | :------------- |
+| roomId | Matrix Room ID |
+
+| Return Code    | Reason     |
+| :------------- | :------------- |
+| 200 | the bridge OKd it and 0 or more results exist |
+| 404 | the room  wasn't found |
+
+### Return Codes
```
lukebarnard1: This header is in the wrong place

config/config.sample.yaml
Half-Shot: I had previously set up a system to only allow listed users the ability to bridge and to control various bits of the bridge. I will eventually be making use of them but I'll probably comment them out for now to avoid confusion.

src/twitter/Timeline.js
Half-Shot: Yeah, I do appear to be missing a !

Provisioning-API.md
```diff
@@ -0,0 +1,102 @@
...
+    //List of strings without #
+    "foo", "bar"
+  ]
+}
+```
+
+## /_matrix/provision/show/timeline/%screenName%
+### Methods
+
+GET Retrive some profile information about a timeline.
```
lukebarnard1: Retrieve

src/twitter/Timeline.js
Half-Shot: Or rather, a ``=== false`` would make more sense

twitter-as.js
Half-Shot: It probably would make more sense to have === undefined in this case, since that's what it returns.

Provisioning-API.md
Half-Shot: Hmm, a 404 doesn't make a lot of sense here. Not sure why I put that in.

lukebarnard1 (approved):
LGTM :)

PR #25, opened 2016-11-21
Allow `user_id` opt in the `manageLink` request body

lukebarnard1:
For the link endpoints, optionally allow a user_id, that will only be used if the userId query parameter is not set.


Half-Shot:
LGTM :+1:

PR #28, opened 2016-11-22
Use .bind(this) instead of keeping track of self

lukebarnard1:
Given that there were a few bugs with `self` being used where it should have been `this` and vice-versa, I thought it might be a good idea to just use `.bind(this)` on those coroutinised functions.


Half-Shot:
Didn't know you could do this, neat :smile:

Half-Shot:
LGTM

PR #29, opened 2016-11-22
Fix provisioning unlink

lukebarnard1:
Previously, the twitter API was being hit for every timeline linked to a room, and would also pass the twitter ID instead of the screenname.

### DIFF
--- a/src/Provisioner.js
+++ b/src/Provisioner.js
@@ -202,7 +202,7 @@ class Provisioner {
     }
   }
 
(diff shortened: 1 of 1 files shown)


Half-Shot:
Hmm, why did you put self back? I thought bind took care of this?

lukebarnard1:
OK, so it turns out there were other issues with unlink. I'll make a new PR that is a bit more sane and with  even less "self".

PR #30, opened 2016-11-23
Fix the unlink endpoint, mainly for timelines

lukebarnard1:
Previously, the twitter API was being hit for every timeline linked to a room, and would also pass the twitter ID instead of the `screenName`.

Also, `remove_timeline` was being given a `screenName` when it should be given the id of the profile to remove. This is now fixed.

(See #29 for previous attempt at making a PR)


Half-Shot:
(Failing test appears to be racing, :disappointed:) 
Thanks LGTM

PR #39, opened 2016-12-07
(Untested) Handle stream creation errors separately to stream errors

lukebarnard1:
Instead of trying to re-attach a twitter user stream when an error happens on attaching, log the error. Only try to re-attach with backoff when a stream error occurs.


Half-Shot:
Looks good. I'll test it later today.

Half-Shot:
Okay. Merging.

PR #41, opened 2016-12-14
Initial commit for enabling OAuth via redirect, as opposed to PIN code

lukebarnard1:


Half-Shot:
LGTM :dog:

I'll have to find a way to trim the file down a bit though, it's getting a bit hefty.

Issue #46, opened 2016-12-16
OAuth tokens are cached even after unOAuth


lukebarnard1:
To reproduce with one Matrix user ID (using riot.im/develop -> room settings -> Manage Integrations -> Twitter)
 1. OAuth using twitter account A using the "Sign in with Twitter" button on Riot, get a new timeline room for A
 2. "Remove Credentials"
 3. Repeat 1. but with twitter account B
 4. See timeline tweets from account A being sent into the new timeline room B

This is not necessarily vulnerable to attack, because OAuth has to be done for account A in the first place. But now I can't use my (account B's) timeline room because it will presumably send tweets as account A.

lukebarnard1:
Restarting the bridge fixes this so presumably the handling around some inner cache is failing and/or unOAuth needs to blow away some Twitter client state.

The cache in question is probably [in TwitterClientFactory](https://github.com/Half-Shot/matrix-appservice-twitter/blob/develop/src/twitter/TwitterClientFactory.js#L147). There _is_ a timeout, set to 10mins [here](https://github.com/Half-Shot/matrix-appservice-twitter/blob/develop/src/twitter/TwitterClientFactory.js#L8). This means that if the client was last verified more than 10 minutes ago, it will be re-verified. 

However, if the user had a twitter client still in the cache (which never get's blown out), the bridge will check that the credentials for the old client are still valid (which of course they are, because no one told Twitter to revoke the creds).

I propose that the cache is invalidated when the twitter account is removed, and the Twitter API is made aware of the credentials being discarded.

PR #51, opened 2016-12-19
Invalidate internal cache on unOAuth

lukebarnard1:
This fixes the case experienced when a previous twitter client is cached for a given user_id who has successfully unOAuthed, and then re OAuths only to see that the previous twitter client is still cached for them.

(Fixes https://github.com/Half-Shot/matrix-appservice-twitter/issues/46)


lukebarnard1:
FTR, the tests are failing due to linting errors (outside of this PR)

Half-Shot:
Yup LGTM. Merging.

PR #60, opened 2017-01-10
Fix to use correct user id when uploading media

lukebarnard1:
Instead of recreating user IDs, reuse the one used for posting messages.


Half-Shot:
LGTM

PR #61, opened 2017-01-10
Fix unhandled exception when expanding URLs

lukebarnard1:


lukebarnard1:
This doesn't actually work, bare with

PR #64, opened 2017-01-12
Add config timelines.shouldSyncInitially

lukebarnard1:
Prevent 100 tweets per timeline being polled for on startup if this is set to false. Default to true for backwards compat.

If set to false, the first poll for a timeline (on startup `tline.hasProcessedTweets = false`, on provision or DB deletion, since = null) will be done with `req.count = 1`, so that only one tweet is received. This tweet is the most recent tweet and is not bridged, but it's `since` is stored so that future polls will be restricted to tweets that were created after startup.

### DIFF
--- a/src/twitter/Timeline.js
+++ b/src/twitter/Timeline.js
@@ -26,6 +26,10 @@ class Timeline {
       timelines: cfg_timelines,
       hashtags: cfg_hashtags
     }
+    // Set default
+    if (this.config.timelines.shouldSyncInitially === undefined) {
+      this.config.timelines.shouldSyncInitially = true;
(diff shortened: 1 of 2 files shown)


Half-Shot:
Hmm, does this make the ``is_new`` system redundant? From what I'm reading, it does?

lukebarnard1:
Ah... well apparently when the bridge starts up, `is_new` is not set to true when the timelines are added.

Would it be a good idea to instead set `is_new = true` in https://github.com/Half-Shot/matrix-appservice-twitter/blob/develop/twitter-as.js#L155 ?

PR #65, opened 2017-01-13
Use the new form of logging with tests

lukebarnard1:
Tests were complaining about not having `npmlog`, but it was never used anyway. .src/logging.js was also complaining about it not being initialised.


Half-Shot:
LGTM

PR #66, opened 2017-01-13
Remove injection of matrix-js-sk 0.5.3

lukebarnard1:
This was being injected into the bridge library, but this is now unecessary.


Half-Shot:
Thanks, LGTM

PR #82, opened 2017-01-31
Fix logging without context

lukebarnard1:
Some log lines do not use a context: detect non-context logs as those where the first argument does not contain %

### DIFF
--- a/src/logging.js
+++ b/src/logging.js
@@ -45,9 +45,17 @@ function handle (level, args) {
   if (!log) {
     console.error("Log not initialised"); // eslint-disable-line no-console
   }
-  // Take the first arg as context
-  // apply util.format to the second arg, with the remaining args as input
-  const message = args[0] + " " + util.format.apply(null, args.slice(1));
+
+  // If the first arg does NOT contain '%', assume it is context and append it onto the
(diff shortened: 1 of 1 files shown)


src/logging.js
```diff
@@ -45,9 +45,17 @@ function handle (level, args) {
   if (!log) {
     console.error("Log not initialised"); // eslint-disable-line no-console
   }
-  // Take the first arg as context
-  // apply util.format to the second arg, with the remaining args as input
-  const message = args[0] + " " + util.format.apply(null, args.slice(1));
+
+  // If the first arg does NOT contain '%', assume it is context and append it onto the
```
Half-Shot, line 49: Should we be checking for args.length >= 2 first? If you've got a simple log line, that could be picked up as a context with no message.

Half-Shot:
Everything else LGTM

src/logging.js
lukebarnard1: If by simple, you mean a string with no '%s's, the context will get set to that and then `args = []` and `util.format` will return an empty string. So the logged message is still correct.

Half-Shot: Ah yeah, I see it now. Okay then that's fine.

Half-Shot (approved):

PR #83, opened 2017-02-02
Check for empty rooms before polling tweets.

Half-Shot:
Fixed issue where the feed queue would get stuck if no tweets were found.

---

This is one half of my commits to lower the amount of traffic going into synapse. This will (optionally) disable feeds which have no active real users. Essentially if all the rooms a feed is used in has no real users joined, we skip the feed.

**March edition** : The other half is also included, which checks too see if a huge member event spam attack is likely and forces the bridge to use a single user.

### DIFF
--- a/src/TweetProcessor.js
+++ b/src/TweetProcessor.js
@@ -72,8 +72,8 @@ class TweetProcessor {
    * This function will fill the content structure for a new matrix message for a given tweet.
    *
    * @param  {TwitterTweet} tweet The tweet object from the Twitter API See {@link}
-   * @param  {type} type  The 'msgtype' of a message.
-   * @return {object}     The content of a Matrix 'm.room.message' event.
+   * @param  {Type} type  The 'msgtype' of a message.
+   * @return {Object}     The content of a Matrix 'm.room.message' event.
    *
    * @see {@link https://dev.twitter.com/overview/api/tweets}
    */
@@ -202,15 +202,13 @@ class TweetProcessor {
 
 
   /**
-   * A function used to process tweets and post them to rooms.
-   *
-   * @param  {type} rooms Rooms to post the tweets to.
-   * @param  {type} tweets Tweets to process
-   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
-   * replies) to be traversed. Set this to how deep you wish to traverse and it
-   * will be decreased when the function calls itself.
-   * @param  {type} opts.client = null The twitter authed client to use.
-   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
+   * Simliar to process_tweet but returns a promise for when all tweets have
+   * been sent. This function is intended to optimise the processing of
+   * tweets in a given array.
+   * @param  {String} rooms
+   * @param  {TwitterTweet[]} tweets
+   * @param  {Object} opts
+   * @see this.process_tweet()
    */
   process_tweets (rooms, tweets, opts) {
     if (opts == null) {
@@ -230,15 +228,22 @@ class TweetProcessor {
   }
 
   /**
-   * A function used to process a tweet and post them to rooms.
+   * Process a given tweet (including esolving any parent tweets),
+   * and submit it to the given room. This function is recursive,
+   * limited to the depth set.
    *
-   * @param  {type} rooms Rooms to post the tweets to.
-   * @param  {type} tweet Tweet to process
-   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
+   * @param  {String} rooms Matrix Room ID of the room that we are processing.
+   * @param  {TwitterTweet} tweet The tweet object from the Twitter API See {@link}
+   * @param  {Object} opts Options for the processor.
+   * @param  {Number} opts.depth The maximum depth of the tweet chain (replies to
    * replies) to be traversed. Set this to how deep you wish to traverse and it
    * will be decreased when the function calls itself.
-   * @param  {type} opts.client = null The twitter authed client to use.
-   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
+   * @param  {TwitterClient} opts.client = null The twitter authed client to use.
+   * @param  {String} opts.force_user_id Should we force one account to post for the tweet.
+   * @return {Promise}  A promise that resolves once the tweet has been queued.
+   *
+   * @see {@link https://dev.twitter.com/overview/api/tweets}
+   */
    */
   process_tweet (rooms, tweet, opts) {
     if (opts == null) {
@@ -254,20 +259,7 @@ class TweetProcessor {
   }
 
   /**
-   * TweetProcessor.prototype._process_tweet - Process a given tweet (including
-   * resolving any parent tweets), and submit it to the given room. This function
-   * is recursive, limited to the depth set.
-   *
-   * @param  {String} rooms Matrix Room ID of the room that we are processing.
-   * @param  {TwitterTweet} tweet The tweet object from the Twitter API See {@link}
-   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
-   * replies) to be traversed. Set this to how deep you wish to traverse and it
-   * will be decreased when the function calls itself.
-   * @param  {type} opts.client = null The twitter authed client to use.
-   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
-   * @return {Promise[]]}  A promise that resolves once the tweet has been queued.
-   *
-   * @see {@link https://dev.twitter.com/overview/api/tweets}
+   * @see this.process_tweet()
    */
   _process_tweet (rooms, tweet, depth, opts) {
     depth--;
--- a/src/TweetProcessor.js
+++ b/src/TweetProcessor.js
@@ -7,6 +7,7 @@ const Promise = require('bluebird');
 
 const TWITTER_MSG_QUEUE_INTERVAL_MS = 150;
 const MSG_QUEUE_LAGGING_THRESHOLD = 50; // The number of messages to be stored in the msg queue before we complain about lag.
+const DEFAULT_TWEET_DEPTH = 1;
 
 class TweetProcessor {
   constructor (opts) {
@@ -54,7 +55,7 @@ class TweetProcessor {
       }
       const msgs = this.msg_queue.pop();
       for(const msg of msgs) {
-        const intent = this._bridge.getIntent(msg.userId);
+        const intent = this._bridge.getIntentFromLocalpart(msg.userId);
         promises.push(intent.sendEvent(msg.roomId, msg.type, msg.content).then(res => {
           if (msg.content.msgtype === "m.text" ) {
             this._storage.add_event(res.event_id, msg.userId, msg.roomId, msg.content.tweet_id, Date.now());
@@ -76,7 +77,7 @@ class TweetProcessor {
    *
    * @see {@link https://dev.twitter.com/overview/api/tweets}
    */
-  tweet_to_matrix_content (tweet, type) {
+  tweet_to_matrix_content (tweet, type, on_behalf_of) {
     let text = tweet.full_text || tweet.text;
     let tags = [];
     if (!tweet.entities) {
@@ -94,6 +95,9 @@ class TweetProcessor {
       })
     }
     text = HTMLDecoder.decode(text);
+    if (on_behalf_of) {
+      text = `@${tweet.user.screen_name}: ${text}`;
+    }
 
     const mxtweet = {
       "body": text,
@@ -110,6 +114,10 @@ class TweetProcessor {
       mxtweet.retweet = tweet._retweet_info;
     }
 
+    if (on_behalf_of) {
+      mxtweet.on_behalf_of = on_behalf_of;
+    }
+
     return mxtweet;
   }
 
@@ -124,11 +132,11 @@ class TweetProcessor {
     return text;
   }
 
-  _push_to_msg_queue (muser, roomid, tweet, type) {
+  _push_to_msg_queue (muser, roomid, tweet, type, on_behalf_of) {
     const time = Date.parse(tweet.created_at);
     let content;
     try {
-      content = this.tweet_to_matrix_content(tweet, type)
+      content = this.tweet_to_matrix_content(tweet, type, on_behalf_of);
     } catch (e) {
       return Promise.reject("Tweet was missing user field.", e);
     }
@@ -194,29 +202,55 @@ class TweetProcessor {
 
 
   /**
-   * A function used to process
+   * A function used to process tweets and post them to rooms.
    *
-   * @param  {type} roomid        description
-   * @param  {type} tweets        description
-   * @param  {type} depth         description
-   * @param  {type} client = null description
-   * @return {type}               description
+   * @param  {type} rooms Rooms to post the tweets to.
+   * @param  {type} tweets Tweets to process
+   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
+   * replies) to be traversed. Set this to how deep you wish to traverse and it
+   * will be decreased when the function calls itself.
+   * @param  {type} opts.client = null The twitter authed client to use.
+   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
    */
-  process_tweets (rooms, tweets, depth, client = null) {
-    if (client == null) {
-      client = this._tclient;
+  process_tweets (rooms, tweets, opts) {
+    if (opts == null) {
+      opts = { }
     }
+    if (opts.client == null) {
+      opts.client = this._tclient;
+    }
+    if (opts.depth == null) {
+      opts.depth = DEFAULT_TWEET_DEPTH;
+    }
+    const promises = [];
     tweets.forEach( (tweet) => {
-      this._process_tweet(rooms, tweet, depth, client);
+      promises.push(this._process_tweet(rooms, tweet, opts.depth, opts));
     });
-
+    return Promise.all(promises);
   }
 
-  process_tweet (rooms, tweet, depth, client = null) {
-    if (client == null) {
-      client = this._tclient;
+  /**
+   * A function used to process a tweet and post them to rooms.
+   *
+   * @param  {type} rooms Rooms to post the tweets to.
+   * @param  {type} tweet Tweet to process
+   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
+   * replies) to be traversed. Set this to how deep you wish to traverse and it
+   * will be decreased when the function calls itself.
+   * @param  {type} opts.client = null The twitter authed client to use.
+   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
+   */
+  process_tweet (rooms, tweet, opts) {
+    if (opts == null) {
+      opts = { }
     }
-    this._process_tweet(rooms, tweet, depth, client);
+    if (opts.client == null) {
+      opts.client = this._tclient;
+    }
+    if (opts.depth == null) {
+      opts.depth = DEFAULT_TWEET_DEPTH;
+    }
+    return this._process_tweet(rooms, tweet, opts.depth, opts);
   }
 
   /**
@@ -226,21 +260,23 @@ class TweetProcessor {
    *
    * @param  {String} rooms Matrix Room ID of the room that we are processing.
    * @param  {TwitterTweet} tweet The tweet object from the Twitter API See {@link}
-   * @param  {Number} depth  The maximum depth of the tweet chain (replies to
+   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
    * replies) to be traversed. Set this to how deep you wish to traverse and it
    * will be decreased when the function calls itself.
+   * @param  {type} opts.client = null The twitter authed client to use.
+   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
    * @return {Promise[]]}  A promise that resolves once the tweet has been queued.
    *
    * @see {@link https://dev.twitter.com/overview/api/tweets}
    */
-  _process_tweet (rooms, tweet, depth, client) {
+  _process_tweet (rooms, tweet, depth, opts) {
     depth--;
     const type = "m.text";
     let promise;
     if (tweet.in_reply_to_status_id_str != null && depth > 0) {
-      promise = client.get('statuses/show/' + tweet.in_reply_to_status_id_str, {})
+      promise = opts.client.get('statuses/show/' + tweet.in_reply_to_status_id_str, {})
       .then((newtweet) => {
-        return this._process_tweet(rooms, newtweet, depth, client);
+        return this._process_tweet(rooms, newtweet, depth, opts);
       }).catch(error => {
         log.error("process_tweet: GET /statuses/show returned: " + error);
         throw error;
@@ -257,7 +293,7 @@ class TweetProcessor {
       promise = promise.then(() => { return this._twitter.profile.update(tweet.user) });
     }
 
-    promise.then( () => {
+    return promise.then( () => {
       if(typeof rooms == "string") {
         rooms = [rooms];
       }
@@ -265,15 +301,16 @@ class TweetProcessor {
         this._storage.room_has_tweet(roomid, tweet.id_str).then(
           (room_has_tweet) => {
             if (!room_has_tweet) {
+              const realUserId = '_twitter_'+tweet.user.id_str;
+              const userId = opts.force_user_id == null ? realUserId : opts.force_user_id
               this._push_to_msg_queue(
-                '@_twitter_'+tweet.user.id_str + ':' + this._bridge.opts.domain, roomid, tweet, type
+                userId, roomid, tweet, type, opts.force_user_id != null ? realUserId : null
               );
             }
           }
         );
       });
     });
-    return promise;
   }
 }
 
--- a/src/TweetProcessor.js
+++ b/src/TweetProcessor.js
@@ -244,7 +244,6 @@ class TweetProcessor {
    *
    * @see {@link https://dev.twitter.com/overview/api/tweets}
    */
-   */
   process_tweet (rooms, tweet, opts) {
     if (opts == null) {
       opts = { }
@@ -289,16 +288,17 @@ class TweetProcessor {
       if(typeof rooms == "string") {
         rooms = [rooms];
       }
-      rooms.forEach((roomid) => {
-        this._storage.room_has_tweet(roomid, tweet.id_str).then(
+      return rooms.map((roomid) => {
+        return this._storage.room_has_tweet(roomid, tweet.id_str).then(
           (room_has_tweet) => {
             if (!room_has_tweet) {
               const realUserId = '_twitter_'+tweet.user.id_str;
               const userId = opts.force_user_id == null ? realUserId : opts.force_user_id
-              this._push_to_msg_queue(
+              return this._push_to_msg_queue(
                 userId, roomid, tweet, type, opts.force_user_id != null ? realUserId : null
               );
             }
+            return Promise.resolve();
           }
         );
       });
--- a/src/TweetProcessor.js
+++ b/src/TweetProcessor.js
@@ -240,7 +240,7 @@ class TweetProcessor {
    * will be decreased when the function calls itself.
    * @param  {TwitterClient} opts.client = null The twitter authed client to use.
    * @param  {String} opts.forceUserId Should we force one account to post for the tweet.
-   * @param  {String} opts.hotRooms Which rooms should be forced to use force_user_id.
+   * @param  {Set<string>} opts.hotRooms Which rooms should be forced to use force_user_id.
    * @return {Promise}  A promise that resolves once the tweet has been queued.
    *
    * @see {@link https://dev.twitter.com/overview/api/tweets}
@@ -289,10 +289,12 @@ class TweetProcessor {
         rooms = [rooms];
       }
       return [...rooms].map((roomid) => {
+        // Are we limiting joins on this room?
+        const forceUserId = opts.hotRooms.has(roomid) ? opts.forceUserId : null;
         return this._processTweetForRoom(
           roomid,
           tweet,
-          opts.hotRooms.has(roomid) ? opts.forceUserId : null);
+          forceUserId);
       });
     });
   }
--- a/src/TweetProcessor.js
+++ b/src/TweetProcessor.js
@@ -127,7 +127,7 @@ class TweetProcessor {
       const start = offset + url.indices[0];
       const end = offset + url.indices[1];
       text = text.substr(0, start) + url.expanded_url + text.substr(end);
-      offset += url.expanded_url.length - (end-start);
+      offset += url.expanded_url.length - (end - start);
     }
     return text;
   }
@@ -208,7 +208,7 @@ class TweetProcessor {
    * @param  {String} rooms
    * @param  {TwitterTweet[]} tweets
    * @param  {Object} opts
-   * @see this.process_tweet()
+   * @see process_tweet()
    */
   process_tweets (rooms, tweets, opts) {
     if (opts == null) {
@@ -258,11 +258,10 @@ class TweetProcessor {
   }
 
   /**
-   * @see this.process_tweet()
+   * @see process_tweet()
    */
   _process_tweet (rooms, tweet, depth, opts) {
     depth--;
-    const type = "m.text";
     let promise;
     if (tweet.in_reply_to_status_id_str != null && depth > 0) {
       promise = opts.client.get('statuses/show/' + tweet.in_reply_to_status_id_str, {})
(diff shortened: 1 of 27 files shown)


src/twitter/Timeline.js
```diff
@@ -17,28 +19,44 @@ class Timeline {
...
+    }
+  }
 
   /**
    * Start the timeline timer so that timelines will be processed in turn.
    */
   start_timeline () {
-    this._t_intervalID = setInterval(() => {this._process_timeline();}, TIMELINE_POLL_INTERVAL);
+    this._t_intervalID = setInterval(() => {
+      Promise.coroutine(this._process_timeline.bind(this))()
```
lukebarnard1: semi-colon

src/twitter/Timeline.js
```diff
@@ -55,7 +73,9 @@ class Timeline {
    * Start the hashtag timer so that hashtags are processed in turn.
    */
   start_hashtag () {
-    this._h_intervalID = setInterval(() => {this._process_hashtags();}, HASHTAG_POLL_INTERVAL);
+    this._h_intervalID = setInterval(() => {
+      Promise.coroutine(this._process_hashtags.bind(this))()
```
lukebarnard1: semi-colon

src/twitter/Timeline.js
```diff
@@ -222,13 +250,27 @@ class Timeline {
...
   }
 
-  _process_timeline () {
+  * _process_timeline () {
     if (this._timelines.length === 0) {
       return;
     }
+    this._t++;
+    if(this._t >= this._timelines.length) {
+      this._t = 0;
```
lukebarnard1: It took me a while to realise what was going on here; better leave a comment to explain things.

src/twitter/Timeline.js
```diff
@@ -222,13 +250,27 @@ class Timeline {
...
+    }
 
-    var tline = this._timelines[this._t];
-    var req = {
+    const tline = this._timelines[this._t];
+    if (!this.config.timelines.poll_if_empty) {
+      if (tline.room.every((roomId) =>{
+        this._empty_rooms.includes(roomId);
```
lukebarnard1: This O(n^2) check scares me a bit. I would just use a for loop here:
```js
let haystack = [1,2,3,4,5,6,7,8,9];

let needles = [1,4,7,9,10]

let needlesInHaystack = true;

for (var i = 0; i < needles.length; i++) {
    if (haystack.indexOf(needles[i]) === -1) {
        needlesInHaystack = false;
        break;
    }
}

if (needlesInHaystack) {
    return;
}
```

src/twitter/Timeline.js
```diff
@@ -240,50 +282,59 @@ class Timeline {
...
-
-    this._t++;
-    if(this._t >= this._timelines.length) {
-      this._t = 0;
     }
   }
 
-  _process_hashtags () {
-    if (this._hashtags.length < 1) {
+  * _process_hashtags () {
```
lukebarnard1: Maintaining near-identical code-paths for hashtags/timelines is going to get really annoying.

src/twitter/Timeline.js
```diff
@@ -240,50 +282,59 @@ class Timeline {
...
+      this._h = 0;
+    }
+
+    const client = yield this.twitter.client_factory.get_client();
+    const feed = this._hashtags[this._h];
+
+
+    if (!this.config.hashtags.poll_if_empty) {
+      if (feed.room.every((roomId) =>{
+        return this._empty_rooms.includes(roomId);
```
lukebarnard1: As above (https://github.com/Half-Shot/matrix-appservice-twitter/pull/83/files#r99139733)

src/twitter/Timeline.js
```diff
@@ -240,50 +282,59 @@ class Timeline {
...
+    }
+    catch (error) {
+      log.error("Timeline", "_process_timeline: GET /statuses/user_timeline returned: %s", error.code);
+      return;
+    }
+    if (feed.length === 0) {
+      return;
+    }
+    else if(feed.length === TIMELINE_TWEET_FETCH_COUNT) {
+      log.info("Timeline", "Poll request hit count limit. Request likely incomplete.");
```
lukebarnard1: Might be worth mention which timeline. I would suggest putting `tline.twitter_id` in logs for `process_timeline` and `process_hashtags`

src/twitter/Timeline.js
lukebarnard1: Also, I would make `_process_timeline` and `_process_hashtags` the same plurality - probably both singular as it is in other places

src/twitter/Timeline.js
```diff
@@ -296,35 +347,49 @@ class Timeline {
...
-    });
+    }
+  }
 
-    this._h++;
-    if(this._h >= this._hashtags.length) {
-      this._h = 0;
+  * _check_empty_rooms () {
+    const bot = this.twitter.bridge.getBot();
+    const memberLists = yield bot.getMemberLists();
```
lukebarnard1: This does a `/sync`, albeit a filtered one. I guess this is OK if it's only being done every 30s.

src/twitter/Timeline.js
```diff
@@ -296,35 +347,49 @@ class Timeline {
...
+    const empty_room_list = [];
+    for (const room in memberLists) {
+      if (memberLists[room].realJoinedUsers.length === 0) {
+        log.silly("Timeline", "%s has no real users", room);
+        empty_room_list.push(room);
+      }
     }
+    this._empty_rooms = empty_room_list;
+    return false;
```
lukebarnard1: Why return false?

src/twitter/Twitter.js
```diff
@@ -31,6 +31,9 @@ class Twitter {
     this._dm = new DirectMessage(this);
     this._config.timelines.poll_if_empty = this._config.timelines.poll_if_empty || false;
     this._config.hashtags.poll_if_empty = this._config.hashtags.poll_if_empty || false;
+    this._config.timelines.poll_if_empty = this._config.timelines.poll_if_empty || false;
+    this._config.hashtags.poll_if_empty = this._config.hashtags.poll_if_empty || false;
+
```
lukebarnard1: The above additions are redundant

lukebarnard1 (changes requested):

src/twitter/Timeline.js
```diff
@@ -1,7 +1,9 @@
 const log      = require('../logging.js');
+const Promise  = require('bluebird');
 
 const TIMELINE_POLL_INTERVAL = 3010; //Twitter allows 300 calls per 15 minute (We add 10 milliseconds for a little safety).
 const HASHTAG_POLL_INTERVAL = 2010; //Twitter allows 450 calls per 15 minute (We add 10 milliseconds for a little safety).
+const EMPTY_ROOM_INTERVAL = 30000;
```
lukebarnard1: This is 30s, not 5 mins as it says in #5

src/TweetProcessor.js
```diff
@@ -194,29 +202,55 @@ class TweetProcessor {
...
+    return Promise.all(promises);
   }
 
-  process_tweet (rooms, tweet, depth, client = null) {
-    if (client == null) {
-      client = this._tclient;
+  /**
+   * A function used to process a tweet and post them to rooms.
+   *
+   * @param  {type} rooms Rooms to post the tweets to.
```
lukebarnard1: *tweet

src/TweetProcessor.js
```diff
@@ -194,29 +202,55 @@ class TweetProcessor {
...
     });
-
+    return Promise.all(promises);
   }
 
-  process_tweet (rooms, tweet, depth, client = null) {
-    if (client == null) {
-      client = this._tclient;
+  /**
+   * A function used to process a tweet and post them to rooms.
```
lukebarnard1: *post it to rooms

src/TweetProcessor.js
```diff
@@ -194,29 +202,55 @@ class TweetProcessor {
...
+   *
+   * @param  {type} rooms Rooms to post the tweets to.
+   * @param  {type} tweet Tweet to process
+   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
+   * replies) to be traversed. Set this to how deep you wish to traverse and it
+   * will be decreased when the function calls itself.
+   * @param  {type} opts.client = null The twitter authed client to use.
+   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
```
lukebarnard1: *for the tweet

src/TweetProcessor.js
```diff
@@ -226,21 +260,23 @@ class TweetProcessor {
    *
    * @param  {String} rooms Matrix Room ID of the room that we are processing.
    * @param  {TwitterTweet} tweet The tweet object from the Twitter API See {@link}
-   * @param  {Number} depth  The maximum depth of the tweet chain (replies to
+   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
```
lukebarnard1: What's wrong with `Number`?

src/TweetProcessor.js
```diff
@@ -226,21 +260,23 @@ class TweetProcessor {
...
    * replies) to be traversed. Set this to how deep you wish to traverse and it
    * will be decreased when the function calls itself.
+   * @param  {type} opts.client = null The twitter authed client to use.
+   * @param  {string} opts.force_user_id Should we force one account to post for every tweet.
```
lukebarnard1: Should this be `string` or `String`? Either way it's good to stay consistent.

src/TweetProcessor.js
lukebarnard1: I'm not sure if doc-ing the same thing three times is the most maintainable or readable way to do things. I would probably document one of them.

Also, I've mentioned it for a couple of arguments but check your argument types, most of them are just type.

src/TweetProcessor.js
```diff
@@ -257,23 +293,24 @@ class TweetProcessor {
...
-    promise.then( () => {
+    return promise.then( () => {
       if(typeof rooms == "string") {
         rooms = [rooms];
       }
       rooms.forEach((roomid) => {
         this._storage.room_has_tweet(roomid, tweet.id_str).then(
           (room_has_tweet) => {
             if (!room_has_tweet) {
+              const realUserId = '_twitter_'+tweet.user.id_str;
```
lukebarnard1: ~spaces around the "+" would be nice~

actually, it looks like this is the style that you're using elsewhere unless I'm mistaken

src/TweetProcessor.js
```diff
@@ -257,23 +293,24 @@ class TweetProcessor {
...
+              const userId = opts.force_user_id == null ? realUserId : opts.force_user_id
               this._push_to_msg_queue(
-                '@_twitter_'+tweet.user.id_str + ':' + this._bridge.opts.domain, roomid, tweet, type
+                userId, roomid, tweet, type, opts.force_user_id != null ? realUserId : null
```
lukebarnard1: `==` is only useful (imho) on line 305 if you intend on allowing `opts.force_user_id` to be the empty string. Using the `opts.force_user_id` itself would allow you to prevent the empty string from being used (as it coerces to false).

I think it'd be clearer to take that last argument and put it in a `const on_behalf_of`, just for clarity. Also, the negation on 307 is screwing with my brain. 

It looks like you want `on_behalf_of` set to `realUserId` when `opts.force_user_id`, and otherwise set it to null.

So overall
```js
let on_behalf_of = null;
let user_id = realUserId;
if (opts.force_user_id) {
    userId = opts.force_user_id;
    on_behalf_of = realUserId;
}
...
```
would be clearer I think

src/TweetProcessor.js
```diff
@@ -257,23 +293,24 @@ class TweetProcessor {
...
       if(typeof rooms == "string") {
         rooms = [rooms];
       }
       rooms.forEach((roomid) => {
         this._storage.room_has_tweet(roomid, tweet.id_str).then(
           (room_has_tweet) => {
             if (!room_has_tweet) {
+              const realUserId = '_twitter_'+tweet.user.id_str;
+              const userId = opts.force_user_id == null ? realUserId : opts.force_user_id
               this._push_to_msg_queue(
```
lukebarnard1: The promise returned by `_push_msg_queue` is being dropped. I'd suggest using `map` on line 300 and then use `Promise.all` to wrap those promises, and return a promise in the anonymous function declared on line 296.

src/handlers/AccountServices.js
```diff
@@ -59,7 +59,7 @@ class AccountServices {
       backoff: 2,
       max_tries: RETRY_INVITE_COUNT
     }).then( () => {
-      var rroom = new RemoteRoom("service_"+event.sender);
+      const rroom = new RemoteRoom("service_"+event.sender);
```
lukebarnard1: I'm not sure about `rroom` as a variable name. I'd prefer `remoteRoom`... Although it seems you haven't decided which case to use for variables in the bridge: some are `some_var`, some are `svar`, some are `somevar`, some are `someVar`.

src/handlers/AccountServices.js
```diff
@@ -411,15 +411,15 @@ ${dm_rooms}`
...
+              const data = {
                 oauth_token: oAuthToken,
                 oauth_secret: oAuthTokenSecret,
                 access_token: null,
                 access_token_secret: null,
                 access_type: access_type
               };
               this._storage.set_twitter_account(id, "", data).then(()=>{
-                var authURL = 'https://twitter.com/oauth/authenticate?oauth_token=' + oAuthToken;
+                const authURL = 'https://twitter.com/oauth/authenticate?oauth_token=' + oAuthToken;
```
lukebarnard1, line 422: spaces around the "+" (?)

src/twitter/Timeline.js
```diff
@@ -1,10 +1,14 @@
 const log      = require('../logging.js');
+const Promise  = require('bluebird');
+const Util  = require('../util.js');
 
 const TIMELINE_POLL_INTERVAL = 3010; //Twitter allows 300 calls per 15 minute (We add 10 milliseconds for a little safety).
 const HASHTAG_POLL_INTERVAL = 2010; //Twitter allows 450 calls per 15 minute (We add 10 milliseconds for a little safety).
+const EMPTY_ROOM_INTERVAL = 30000;
```
lukebarnard1: Probably a good idea to put `_MS` on the end of these intervals

src/twitter/Timeline.js
```diff
@@ -1,10 +1,14 @@
 const log      = require('../logging.js');
+const Promise  = require('bluebird');
+const Util  = require('../util.js');
```
lukebarnard1: inconsistent spacing

src/twitter/Timeline.js
```diff
@@ -65,6 +91,8 @@ class Timeline {
     if (this._h_intervalID) {
       clearInterval(this._h_intervalID);
       this._h_intervalID = null;
+    } else {
+      throw Error("Timer not started");
```
lukebarnard1: This error looks very much like the error in `stop_timeline`. It might be best to differentiate them.

src/twitter/Timeline.js
```diff
@@ -97,40 +132,46 @@ class Timeline {
...
-      obj = {hashtag, room: [] }
+      obj = {hashtag, room: new Set() }
       if(opts.is_new) {
         this._newtags.add("#"+hashtag);
       }
     }
-    if(!obj.room.includes(room_id)) {
-      obj.room.push(room_id);
-    }
+    obj.room.add(room_id);
```
lukebarnard1: `obj` might be `undefined` at this point, if `_find_hashtag` returns `-1`.

src/twitter/Timeline.js
```diff
@@ -97,40 +132,46 @@ class Timeline {
...
     log.info("Added Hashtag: %s", hashtag);
+    return true;
   }
 
   /**
    * Add a Twitters user's timeline to the timeline processor. Tweets will be
    * automatically send to the given room.
    *
+   * HTM* - Use a single account to send tweets, avoiding large numbers of join
+   * events. This is enabled automatically unless disabled via config
```
lukebarnard1: I'm not sure mentioning this here is necessary. I _would_ mention that by default, `opts.high_traffic_mode` is set to false.

src/twitter/Timeline.js
```diff
@@ -97,40 +132,46 @@ class Timeline {
...
+  */
   add_timeline (twitter_id, room_id, opts) {
-    var tline = this._find_timeline(twitter_id);
-    var obj;
-
     if (this.config.timelines.enable === false) {
-      return;
+      return false;
+    }
+    if (!Util.isRoomId(room_id)) {
```
lukebarnard1: Looking at the code paths for this, it's being done _quite_ late. I think the room ID should be checked way before the bridge adds a timeline. In general it's Good Practice:tm: to validate as soon as you have data that has been given by a user that might be malformed or malicious.

src/twitter/Timeline.js
```diff
@@ -144,25 +185,33 @@ class Timeline {
       opts.exclude_replies = false;
     }
 
+    if(opts.high_traffic_mode === undefined) {
```
lukebarnard1: As far as I can see, this is always true. The option is never passed to `add_timeline`.

src/twitter/Timeline.js
```diff
@@ -97,40 +132,46 @@ class Timeline {
...
-
     if (this.config.timelines.enable === false) {
-      return;
+      return false;
+    }
+    if (!Util.isRoomId(room_id)) {
+      throw Error("Not a valid room_id");
     }
+    const tline = this._find_timeline(twitter_id);
+    let obj;
```
lukebarnard1: This needs a better name. It looks like an object containing data pertaining to a timeline, so perhaps `timeline`?

src/twitter/Timeline.js
```diff
@@ -97,40 +132,46 @@ class Timeline {
...
-    var obj;
-
     if (this.config.timelines.enable === false) {
-      return;
+      return false;
+    }
+    if (!Util.isRoomId(room_id)) {
+      throw Error("Not a valid room_id");
     }
+    const tline = this._find_timeline(twitter_id);
```
lukebarnard1: Not obvious to be that this is an index.

src/twitter/Timeline.js
```diff
@@ -171,9 +220,10 @@ class Timeline {
    *
    * @param  {string} twitter_id The twitter id of the timeline to remove.
    * @param  {string} [room_id] If specified, only remove from this room.
+   * @return {boolean} Was the operation successful.
    */
   remove_timeline (twitter_id, room_id) {
-    this._remove_from_queue(true, twitter_id, room_id);
+    return this._remove_from_queue(true, twitter_id, room_id);
```
lukebarnard1: Not sure why this or `remove_hashtag` is returning anything here. Can you clarify?

src/twitter/Timeline.js
```diff
@@ -182,31 +232,30 @@ class Timeline {
...
    * @param  {string} hashtag The hashtag to remove. (without the #)
    * @param  {string} [room_id] If specified, only remove from this room.
+   * @return {boolean} Was the operation successful.
    */
   remove_hashtag (hashtag, room_id) {
-    this._remove_from_queue(false, hashtag, room_id);
+    return this._remove_from_queue(false, hashtag, room_id);
   }
 
   _remove_from_queue (isTimeline, id, room_id) {
```
lukebarnard1: This function needs more comments, I cannot work out what's going on here.

src/twitter/Timeline.js
```diff
@@ -182,31 +232,30 @@ class Timeline {
...
     const i = isTimeline ? this._find_timeline(id) : this._find_hashtag(id);
-    var queue = isTimeline ? this._timelines : this._hashtags;
+    const queue = isTimeline ? this._timelines : this._hashtags;
     if(i !== -1) {
```
lukebarnard1: I would return early here instead of having many lines indented.

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
-      count: TIMELINE_TWEET_FETCH_COUNT,
-      exclude_replies: tline.exclude_replies,
-      tweet_mode: "extended" // https://github.com/Half-Shot/matrix-appservice-twitter/issues/31
-    };
-
-    if(this._newtags.has(req.user_id)) {
-      req.count = 1;
-      this._newtags.delete(req.user_id);
+  is_feed_exceeding_user_limit (tweets, timeline = true) {
+    let max = (timeline ? (TIMELINE_POLL_INTERVAL*this._t) : (HASHTAG_POLL_INTERVAL*this._h))/60000;
```
lukebarnard1: This needs a rename and a comment. `max` really doesn't tell the reader enough.

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
+    max = Math.max(max * NEW_PROFILE_THRESHOLD_MIN, 1)
```
lukebarnard1: This is also quite dubious, pls comment the reason you are doing this. 

Could you explain (here) what is going on as well.

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
+    if ((timeline ? this.config.timelines : this.config.hashtags).single_account_fallback === true) {
+      const user_ids = new Set(tweets.map((tweet) => {tweet.id_str})).size;
+      log.verbose(`is_feed_exceeding_user_limit: ${user_ids} > ${max}`);
```
lukebarnard1: It might be more useful to log verbosely if the feed _is_ exceeding the user limit. Stating the number of ids and the max doesn't seem useful.

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
   }
 
-  _process_hashtags () {
-    if (this._hashtags.length < 1) {
-      return;
+  _process_hashtag () {
+    if (this._hashtags.length === 0) {
+      return Promise.resolve();
     }
+    this._h++;
```
lukebarnard1: It's unclear what `_h` is. It needs a rename.

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
-    }).catch((err) => {
-      log.error("Timeline", "Error whilst processing timeline %s: %s", tline.twitter_id, err);
-    });
-
+  _process_timeline () {
+    if (this._timelines.length === 0) {
+      return Promise.resolve();
+    }
+    // Rotate to the next timeline.
     this._t++;
```
lukebarnard1: `_t` needs a rename to something more understandable

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
     }
+
+    const tline = this._timelines[this._t];
+    return Promise.coroutine(this._process_feed.bind(this))(true, tline);
   }
 
-  _process_hashtags () {
-    if (this._hashtags.length < 1) {
-      return;
+  _process_hashtag () {
```
lukebarnard1: From what I can see, `_process_hashtags` and `_process_timeline` seem to largely do the same thing. It would be nice to factor this stuff out into a single function, which gets given the `timeline` boolean that seems to be what you use elsewhere.

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
 
-    var feed = this._hashtags[this._h];
-    var req = {
-      q: "%23"+feed.hashtag,
-      result_type: 'recent',
-      count: HASHTAG_TWEET_FETCH_COUNT,
+  *_process_feed (isTimeline, feed) {
+    const client = yield this.twitter.client_factory.get_client();
+    const sinceId = isTimeline ? "@"+feed.twitter_id : feed.hashtag;
+    const getPath = isTimeline ? 'statuses/user_timeline' : 'search/tweets'
```
lukebarnard1: semi-colon

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
+    if(this.is_room_excluded(feed.room)) {
+      log.info("Timeline", `${feed.hashtag} is ignored because the room(s) contains no real members`);
```
lukebarnard1: Why `feed.hashtag`? Should this not be `feed.twitter_id` for timelines?

src/twitter/Timeline.js
```diff
@@ -215,115 +264,141 @@ class Timeline {
...
+      return;
+    }
+    else if(results.length === req.count) {
+      log.info("Timeline", "Poll request hit count limit. Request likely incomplete.");
+    }
+    this.twitter.storage.set_since(sinceId, results[0].id_str);
+    const shouldForceUserId = this.is_feed_exceeding_user_limit(results, isTimeline);
+    let force_user_id;
+    if(shouldForceUserId) {
+      log.verbose("Timeline", `Forcing single user mode for ${feed.hashtag}`);
```
lukebarnard1: See https://github.com/Half-Shot/matrix-appservice-twitter/pull/83/files#r108435319

src/twitter/Twitter.js
```diff
@@ -70,6 +71,15 @@ class Twitter {
...
       }
+      this._userstream.start();
+
+      if (this._config.hashtags.enable && this._config.timelines.enable) {
+        this.timeline.start_empty_room_checker();
+      }
+
+      if (this._config.hashtags.enable && this._config.timelines.enable) {
+        this.timeline.start_empty_room_checker();
+      }
```
lukebarnard1: These 3 lines are repeated

test/handlers/test_Timeline.js
```diff
@@ -0,0 +1,370 @@
...
+        hashtag: "test",
+        room: new Set(["bacon"])
+      });
+      timeline._hashtags.push({
+        hashtag: "test1",
+        room: new Set(["bacon"])
+      });
+      timeline._hashtags.push({
+        hashtag: "test2",
+        room: new Set(["bacon"])
```
lukebarnard1: You must have been _quite_ hungry when writing this :bacon:

src/twitter/Timeline.js
lukebarnard1: Where does `high_traffic_mode` get set? How does it get from config to here?

lukebarnard1 (changes requested):

src/TweetProcessor.js
Half-Shot: > I'm not sure if doc-ing the same thing three times is the most maintainable or readable way to do things. I would probably document one of them.

Yes. I'm starting to regret this. I will document the public facing ones and have a ``see function x() for details``.

src/TweetProcessor.js
Half-Shot: Actually, this is me being lazy and awful because I prefer my operators spaced out. I think a rule is needed.

src/TweetProcessor.js
Half-Shot: cd0ab8e79fbe1028c5e8b271d2e4aa6033ab3eb9 looks about right?

src/handlers/AccountServices.js
Half-Shot: Yes, one of the side effects of messing with multiple projects. Another rule I aught to use.

src/twitter/Timeline.js
Half-Shot: ~~Aaaaand a rule for this one too probably.~~

**EDIT**: Apparently nothing exists for eslint that catches this :/

Half-Shot: >obj might be undefined at this point, if _find_hashtag returns -1.

How so? The if statement catches -1?

src/twitter/Timeline.js
Half-Shot: This looks to be mistakenly old code that I probably wanted to use before I changed the way high traffic mode worked. It's now done upon receiving new tweets so this doesn't need to exist at all.

src/twitter/Timeline.js
Half-Shot: Hmm, I was probably being paranoid here. I don't think any routes to add_timeline accept pure roomIds without checking them.

src/twitter/Timeline.js
Half-Shot: I put it in for one of my tests, if it's a problem I can remove it though.

src/TweetProcessor.js
```diff
@@ -230,15 +228,22 @@ class TweetProcessor {
   }
 
   /**
-   * A function used to process a tweet and post them to rooms.
+   * Process a given tweet (including esolving any parent tweets),
```
lukebarnard1: resolving

src/TweetProcessor.js
```diff
@@ -230,15 +228,22 @@ class TweetProcessor {
...
+   * and submit it to the given room. This function is recursive,
+   * limited to the depth set.
    *
-   * @param  {type} rooms Rooms to post the tweets to.
-   * @param  {type} tweet Tweet to process
-   * @param  {type} opts.depth The maximum depth of the tweet chain (replies to
+   * @param  {String} rooms Matrix Room ID of the room that we are processing.
+   * @param  {TwitterTweet} tweet The tweet object from the Twitter API See {@link}
```
lukebarnard1: is this link empty?

src/TweetProcessor.js
```diff
@@ -289,21 +288,41 @@ class TweetProcessor {
...
+  /**
+   * Process a given tweet for a given room on behalf of  _process_tweet
+   *
+   * @param  {String} roomid The matrix roomid of the room.
+   * @param  {TwitterTweet} tweet  description
+   * @param  {Object} opts   See process_tweet()
+   * @return {Promise}
+   * @see process_tweet()
+   */
+  _process_tweet_for_room (roomid, tweet, opts) {
```
lukebarnard1: You _could_ just pass `opts.force_user_id` instead of the entire `opts` object.

src/twitter/Timeline.js
```diff
@@ -212,63 +202,61 @@ class Timeline {
...
+      queue.splice(i, 1);
     }
+
+    if(isTimeline) {
+      this._timelines = queue;
+    }
+    else{
+      this._hashtags = queue;
+    }
+    return true
```
lukebarnard1: semi colon

lukebarnard1 (changes requested):

lukebarnard1:
This is looking really good for preventing the overloading of synapse with individual timelines and hashtags. I'm not sure this will help in the case where someone has bridged a load of different timelines(/hashtags) into a single room.

src/TweetProcessor.js
```diff
@@ -287,8 +288,11 @@ class TweetProcessor {
       if(typeof rooms == "string") {
         rooms = [rooms];
       }
-      return rooms.map((roomid) => {
-        return this._process_tweet_for_room(roomid, tweet, opts);
+      return [...rooms].map((roomid) => {
```
lukebarnard1: I would leave a comment as to why you need to do `[...rooms]`.

lukebarnard1: Also `roomid` ? Surely `roomId` :)

src/TweetProcessor.js
```diff
@@ -299,26 +303,26 @@ class TweetProcessor {
    *
    * @param  {String} roomid The matrix roomid of the room.
    * @param  {TwitterTweet} tweet  description
-   * @param  {Object} opts   See process_tweet()
+   * @param  {String} force_user_id   See process_tweet()
```
lukebarnard1: `forceUserId`

src/twitter/Timeline.js
```diff
@@ -357,24 +390,32 @@ class Timeline {
       log.info("Timeline", "Poll request hit count limit. Request likely incomplete.");
     }
     this.twitter.storage.set_since(sinceId, results[0].id_str);
-    const shouldForceUserId = this.is_feed_exceeding_user_limit(results, isTimeline);
+    const hotRooms = this.roomsExceedingJoinThreshold(results, feed.room, isTimeline);
```
lukebarnard1: I find it confusing that `feed.room` is actually an array. Perhaps `feed.rooms`?

lukebarnard1: As per #twitter:half-shot.uk, comments about `hotRooms` would be good

src/twitter/Timeline.js
```diff
@@ -25,11 +26,16 @@ class Timeline {
     this._hashtags = [] // {hashtag:string, room:[string]}
     this._newtags = new Set();
     this._empty_rooms = new Set();
+    this._hot_room_counter = new Map();
+    this._members_rooms = new Map(); // [string -> Set]
```
lukebarnard1: would be nice to know what `string` was. Is it a `roomId` ?

src/twitter/Timeline.js
```diff
@@ -269,18 +275,26 @@ class Timeline {
...
-    // Number of new profiles that we can accept in this timespan.
-    const maxNProfiles = Math.ceil((currentPollInterval / 60000) * NEW_PROFILE_THRESHOLD_MIN);
-    const userIds = new Set(tweets.map((tweet) => {tweet.id_str})).size;
-    return userIds > maxNProfiles;
+  roomsExceedingJoinThreshold (tweets, rooms, isTimeline) {
+    if(!(isTimeline ? this.config.timelines : this.config.hashtags).single_account_fallback) {
+      return new Set();
+    }
+    // Find any rooms that exceed the threshold.
+    const result = ([...rooms]).filter((room) => {
```
lukebarnard1: `roomId` perhaps?

src/twitter/Timeline.js
```diff
@@ -269,18 +275,26 @@ class Timeline {
...
+      const currentMembers = this._members_rooms.get(room) || new Set();
```
lukebarnard1: might be slightly nicer to have a constant `EMPTY_SET`, which is set to `new Set()`, as it looks like `currentMembers` isn't modified.

src/twitter/Timeline.js
```diff
@@ -383,6 +424,9 @@ class Timeline {
         this._empty_rooms.add(room);
       } else {
         this._empty_rooms.delete(room);
+        this._members_rooms.set(room.id, memberLists[room].remoteJoinedUsers);
+        log.verbose("Timeline", "Clearing hot_room_counter");
+        this._hot_room_counter.clear();
```
lukebarnard1: this definitely needs a comment

lukebarnard1 (changes requested):

src/TweetProcessor.js
Half-Shot: Bah. Yes!

src/twitter/Timeline.js
```diff
@@ -426,11 +426,15 @@ class Timeline {
...
+        // Room is empty, lets ignore it
         this._empty_rooms.add(room);
       } else {
+        // Room is not empty, delete if it exists
         this._empty_rooms.delete(room);
+        // Store the number of bot users away
         this._members_rooms.set(room.id, memberLists[room].remoteJoinedUsers);
         log.verbose("Timeline", "Clearing hot_room_counter");
+        // Clear the collection of hot rooms and let them reset to 0.
         this._hot_room_counter.clear();
```
lukebarnard1: I'm not sure the counter will be that accurate if it is not read immediately before being cleared.

Half-Shot: *Sorted out in #twitter:half-shot.uk*

PR #84, opened 2017-02-14
Typecheck before checking for context

lukebarnard1:


Half-Shot:
Ooh, thanks for finding this. LGTM

PR #85, opened 2017-02-22
timeline.retweets option

erdnaxeli:
I like to know who have done a retweet. I saw that you add this info in the source message, but it is not shown on Riot. So I propose to add a room option `timeline.retweets` to select how to show retweeted status:

- `root` (the default): show the retweeted status like any other status, this is the current behavior
- `rt`: show the status from the user doing the retweet, with `RT @user_retweeted: blabla`

I've added a column in the `timeline_room` table to save this parameter.

If the `rt` parameter is selected, we need to rewrite the the `tweet.text` field, because the original one can be truncated (and its a badly, truncating even links).

If my not really sure:
- about the fact of getting the parameter from the db for every tweet
- about the names `root` and `rt`

### DIFF
--- a/src/TweetProcessor.js
+++ b/src/TweetProcessor.js
@@ -253,27 +253,44 @@ class TweetProcessor {
     promise = promise.then(() => { return this._twitter.profile.update(tweet.user); });
     if (tweet.retweeted_status) {
       tweet.retweeted_status._retweet_info = { id: tweet.id_str, tweet: tweet.user.id_str };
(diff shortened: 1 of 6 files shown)


Half-Shot:
>about the fact of getting the parameter from the db for every tweet

This is fine. You are unlikely to exceed any disk IO barriers, plus I believe sqlite caches values so it's not a problem.

>about the names root and rt

Hmm. That's a tough one. I'd use "default" and "compact"?

PR #87, opened 2017-03-24
Fix stream deaths

Half-Shot:
This PR fixes stream deaths by reconnecting after a timeout, which according to Twitter we should do after 30s. 

I got fed up with the number of ``var`` statements too so I threw them out too, hence the size.

### DIFF
--- a/src/twitter/UserStream.js
+++ b/src/twitter/UserStream.js
@@ -124,6 +124,7 @@ class UserStream {
 
   detach (user_id) {
     if(this._user_streams.has(user_id)) {
+      this._user_keepalive.delete(user_id);
       this._user_streams.get(user_id).destroy();
       this._user_streams.delete(user_id);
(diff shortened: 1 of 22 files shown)


src/twitter/UserStream.js
```diff
@@ -87,13 +87,12 @@ class UserStream {
...
-          "Got 'end'. %s",
-          response
-        );
-      });
+      // stream.on('end', (response) => {
+      //   log.info(
+      //     "UserStream",
+      //     "Got 'end'. %s"
+      //   );
+      // });
```
lukebarnard1: Should this be commented? Or deleted?

lukebarnard1 (commented):
LGTM apart from question.

src/twitter/UserStream.js
Half-Shot: Deleted really. Will do.

PR #88, opened 2017-03-30
Don't update profiles if none of the rooms use them.

Half-Shot:


src/TweetProcessor.js
```diff
@@ -277,11 +277,21 @@ class TweetProcessor {
...
     }
 
-    promise = promise.then(() => { return this._twitter.profile.update(tweet.user); });
+    // If all the rooms are marked as hot, don't bother with updating the profile.
+    const updateProfiles = ![...rooms].every((roomId) => {return opts.hotRooms.has(roomId)});
+    if (updateProfiles) {
+      promise = promise.then(() => { return this._twitter.profile.update(tweet.user); });
+    }
+    else {
+      promise = Promise.resolve();
```
lukebarnard1: This is a bug because you might end up overwriting the promise created on line 268. I think this `else` is redundant.

src/TweetProcessor.js
```diff
@@ -277,11 +277,21 @@ class TweetProcessor {
       promise = Promise.resolve();
     }
 
-    promise = promise.then(() => { return this._twitter.profile.update(tweet.user); });
+    // If all the rooms are marked as hot, don't bother with updating the profile.
+    const updateProfiles = ![...rooms].every((roomId) => {return opts.hotRooms.has(roomId)});
```
lukebarnard1, line 281: This seems like an edgecase?

Half-Shot: Potentially, but it's a fairly cheap thing to check. The case is usually where you have one or two hot rooms such as a popular hashtag room where we are unlikely to ever use the profile and it would be a bit silly to store it.

lukebarnard1: OK, fair enough

lukebarnard1 (approved):

PR #93, opened 2017-05-03
Fix winston (duplicate lines and term colours)

Half-Shot:


config/config.sample.yaml
```diff
@@ -25,11 +25,34 @@ media:
...
+      # ddd: The weekday (Mon, Tue, ..., Sun).
+    datePattern: "yyyy-MM-dd" # Date pattern to be used with daily.
+  file:
+  rotate:
+    date: false
+    size: false
+    count: 1
+    datePattern:  # Date pattern for files split by date
+  compress:
+  level:
```
lukebarnard1: This and the lines above don't look like they should be here.

src/logging.js
```diff
@@ -1,57 +1,96 @@
...
-        return options.timestamp() + options.level + ' ' + (options.message ? options.message : '') +
-          (options.meta && Object.keys(options.meta).length ? '\n\t' + JSON.stringify(options.meta) : '' );
-      },
+      name: "console",
+      timestamp: funcTimestamp,
+      formatter: winstonFormatter,
       level: loggingConfig.level,
-    };
-    transports.push(new Rotate(logrotateConfig));
+      colorize: true,
```
lukebarnard1: So I'm a bit confused as to why we ask winston to colorize but also use chalk to colorize. FTR winston uses a similar thing to chalk, https://www.npmjs.com/package/colors but requires you `addColors` (https://github.com/winstonjs/winston#using-custom-logging-levels).

src/logging.js
```diff
@@ -1,57 +1,96 @@
...
+    levels: {
+      error: 0,
+      warn: 1,
+      info: 2,
+      verbose: 3,
+      silly: 4,
+    }
   });
+  if (loggingConfig.rotate.size && loggingConfig.rotate.daily) {
+    handle("warn", ["You have enabled both 'size' and 'daily' in your logger config. Size will be ignored."]);
```
lukebarnard1: I think it'd be best to just to log an error and die if both are set. It doesn't make sense to have both configured and might be a source of confusion.

config/config.sample.yaml
```diff
@@ -25,11 +25,34 @@ media:
...
+      # M: The month.
+      # MM: The zero padded month.
+      # d: The day.
+      # dd: The zero padded day.
+      # H: The hour.
+      # HH: The zero padded hour.
+      # m: The minute.
+      # mm: The zero padded minute.
+      # ddd: The weekday (Mon, Tue, ..., Sun).
+    datePattern: "yyyy-MM-dd" # Date pattern to be used with daily.
```
lukebarnard1: I would suggest hard-coding this instead of leaving it in config for one reason: for some ungodly reason, winston applies semantics to `datePattern` to determine what granularity of rotation to have. Which means you could end up with a log that _doesn't_ rotate daily despite the name of the thing.

lukebarnard1 (changes requested):

config/config.sample.yaml
Half-Shot: Well that's impressively dumb of it. I'll hardcode it :/

src/logging.js
lukebarnard1: Answer: we're using a custom formatting function.

@Half-Shot, I would suggest leaving not using `colorize` because that indicates that means we would be using winston's own colouring were we not overriding the format function. 

I think I would have a parallel `winstonColourFormatter` or something that does what `winstonFormatter` does but with a coloured level. Then I'd factor out as much as possible from `winstonFormatter` to avoid duplication of code.

src/logging.js
```diff
@@ -1,57 +1,96 @@
...
+    timestamp: funcTimestamp,
+    formatter: winstonFormatter,
     level: loggingConfig.level,
-  }));
+    maxsize: loggingConfig.rotate.size,
+    maxFiles: loggingConfig.rotate.count,
+    zippedArchive: loggingConfig.compress,
+  }
+  if (loggingConfig.file ) {
+    if (loggingConfig.rotate.daily !== true) {
```
lukebarnard1: Perhaps just use `loggingConfig.rotate.daily` amd swap the if-else around

src/logging.js
```diff
@@ -1,57 +1,94 @@
...
+  warn: "yellow",
+  info: "blue",
+  verbose: "white",
+  silly: "grey",
+};
 
-let log = null;
+function winstonFormatter (options) {
+  let level = options.level;
+  if(options.colorize && chalk.supportsColor) {
```
lukebarnard1: I would suggest leaving not using colorize because that indicates that means we would be using winston's own colouring were we not overriding the format function.

I think I would have a parallel winstonColourFormatter or something that does what winstonFormatter does but with a coloured level. Then I'd factor out as much as possible from winstonFormatter to avoid duplication of code.

src/logging.js
```diff
@@ -1,57 +1,94 @@
...
-  }));
-
+    maxsize: loggingConfig.rotate.size,
+    maxFiles: loggingConfig.rotate.count,
+    zippedArchive: loggingConfig.compress,
+  }
   if (loggingConfig.file) {
-    const logrotateConfig = {
-      file: loggingConfig.file,
+    if (!loggingConfig.rotate.daily) {
```
lukebarnard1: Perhaps just use loggingConfig.rotate.daily amd swap the if-else around

lukebarnard1 (changes requested):

lukebarnard1 (approved):

lukebarnard1:
LGTM :star2:

